PR #13, opened 2016-05-17
RunTask failures

rclark:
When the watcher makes a `runTask` request, the ECS API will respond with a "failure" if there are insufficient resources available in the cluster to run the requested task.

For example, if you've requested to run a task with 1024GB RAM but all of the cluster's RAM is already reserved, the response to the `runTask` request will include a message `RESOURCE:MEMORY`.

What I've done here so far is to:

- adjust the system to return messages to SQS if they were received, and then a task failed to run
- adjust the `tasks.outcome.noop` to return messages immediately, with no visibility timeout, even if the env specifies a desire for exponential backoff. Backoff will only apply to messages returned to the queue as a result of a task that ran and failed, not one that failed to run.

The problem here is that exponential backoff is calculated based on the number of times a message has been received. If messages are being received and dropped back to SQS as a no-op due to resource limitations, then once they are actually processed, if they fail, they may get a far too large timeout applied to them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,5 +8,6 @@ module.exports = {
   notifications: require('./lib/notifications'),
   messages: require('./lib/messages'),
   tasks: require('./lib/tasks'),
-  template: require('./lib/template')
+  template: require('./lib/template'),
+  resources: require('./lib/resources')
 };
--- a/lib/main.js
+++ b/lib/main.js
@@ -114,7 +114,7 @@ module.exports = function(config) {
           tasks.run(env, function(err) {
             if (!err) return next();
 
-            log.error(err);
+            log.warn(err.message);
 
             messages.complete({
               reason: err.message,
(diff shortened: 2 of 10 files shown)


rclark:
In the end, the outcome of this PR is:

- adds `lib/resources.js` to handle resource tracking. On creation, a resources object makes a describeTaskDefinition request to find the required resources for the task to be run. It also begins polling the cluster every 60s to get the ARNs of EC2s in the cluster.

- on each loop, after dealing with messages for any completed tasks, makes a describeContainerInstances request for the current set of EC2s. This provides the resources currently available on each EC2.

- then, tries to determine the number of tasks that there are adequate resources to run. If this is > 0, then SQS is polled and the process continues. If there are inadequate resources, the loop restarts.

- if a runTask operation still encounters a failure due to lack of memory resources, the runTask operation is retried every 5s until it gets placed. Any other failure results in the message being returned to the queue immediately.

PR #37, opened 2016-08-16
add log aggregation function

isiyu:
add log aggregation function to template.js for cloudwatch logs

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -126,7 +126,7 @@ module.exports = function(options) {
   };
 
   if (options.user) user(prefixed, resources, references);
-  if (options.webhook) webhook(prefixed, !!options.webhookKey, resources);
+  if (options.webhook) webhook(prefixed, !!options.webhookKey, resources, references);
 
   resources[prefixed('NotificationTopic')] = {
     Type: 'AWS::SNS::Topic',
@@ -148,7 +148,7 @@ module.exports = function(options) {
     }
   };
 
-  if (options.logAggregationFunction) logAggregator(prefixed, resources, references);
+  if (options.logAggregationFunction) logAggregator(prefixed, resources);
 
   resources[prefixed('Queue')] = {
     Type: 'AWS::SQS::Queue',
--- a/lib/template.js
+++ b/lib/template.js
@@ -59,6 +59,8 @@ var pkg = require(path.resolve(__dirname, '..', 'package.json'));
  * between retries will increase exponentially according to the number of times
  * that the job has been retried. Set this parameter to false in order to avoid
  * this exponential backoff and retry jobs immediately.
+ * @param {string} [options.logAggregationFunction=''] - the ARN of the lambda function
+ * to collect and send logs to sumologic
  * @param {string} [options.mounts=''] - if your worker containers need to mount
  * files or folders from the host EC2 file system, specify those mounts with this
  * parameter. A single mount point can be specified as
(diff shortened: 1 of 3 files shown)


isiyu:
Hey - @rclark I was able to get this working with `reverser` , tested by deploying with this update and was able to see the logs using mbx cli and the --ecs flag. 

This is the reverser[ js template](https://github.com/mapbox/reverser/blob/cfn/cloudformation/reverser.template.js#L16). I saw in the ecs-conex template there's a conditional setup to only add the LogForwarding resource if the 'LogAggregationFunction' is defined, I wasn't sure how to set that up that conditional in the template js here. 

Is this the setup you had in mind in terms of how watchbot stacks would use this function? I wasn't sure how much flexibility to add for this param/resource.

rclark:
In conex, if you provide a stack parameter value, then the stack will build the log aggregator for you.

The catch pulling that back into watchbot is that the function that watchbot gives you to use in your js template does not make any parameters -- it only makes resources. So it can't possibly handle conditionals based on parameter inputs for you.

But it can conditionally add resources based on the options that you provide to `watchbot.template()`. I would think that a log aggregation function, provided by watchbot would look more like:

```js
var myTemplate = {
  Parameters: {
    LogAggregator: {
      "Type": "String",
      "Arn": "The ARN of a Lambda function to send logs to"
    }
  }
};

var watcher = watchbot.template({
  ...options...
  logAggregator: { Ref: 'LogAggregator' }
});

module.exports = watchbot.merge(myTemplate, watcher);
```

If you want aggregation, provide the `logAggregator` option. If you don't want it, don't provide that option to `watchbot.template`.

isiyu:
hey @rclark I think this is ready if you could give it a once over. 

I had a couple cfn questions

If the LogForwarding resource was being defined in the main function. How does setting the Destination Arn [this way ](https://github.com/mapbox/ecs-watchbot/blob/9b4d481022850ff9cef6d8f730ffcc9a5cdcc779/lib/template.js#L408) differ from:
```
DestinationArn: options.logAggregationFunction,
```


In the template example you gave - is the `"Arn": "The ARN of a Lambda function to send logs to"` line necessary?  I wasn't able to get it to work when I tried using this with a stack I was building.

lib/template.js
```diff
@@ -140,6 +142,8 @@ module.exports = function(options) {
     }
   };
 
+  if (options.logAggregationFunction) logAggregator(prefixed, resources);
```
rclark: You can move this up alongside the other conditional additiones (L124ish). Even though the content of this depends on a reference to the LogGroup defined on L138, that doesn't actually have to be there first -- by the time we get to the end of everything we'll have all the resources in there that we need.

lib/template.js
```diff
@@ -396,6 +400,18 @@ module.exports = function(options) {
   };
 };
 
+function logAggregator(prefixed, resources) {
+  resources[prefixed('LogForwarding')] = {
+    Type: 'AWS::Logs::SubscriptionFilter',
+    Description: 'Sends log events from CloudWatch Logs to a Lambda function',
+    Properties: {
+      DestinationArn: cf.ref('LogAggregationFunction'),
```
rclark: Instead of wiring a dependency on a template parameter called LogAggregationFunction, you should be able to drop `options.logAggregationFunction` here. This gives the template writer flexibility to provide that ARN however they see fit.

PR #44, opened 2016-09-03
Add logStream functionality

### LINKED ISSUE #41: watchbot.log should support streaming
Some aspect of `logs.js` should return a writable stream that you can pipe logs into. Maybe like

```
var log = watchbot.logStream();
myChildProcess.stdout.pipe(log);
```

cc @jakepruitt

jakepruitt:
cc/ @rclark 
fixes #41

### DIFF
--- a/lib/logs.js
+++ b/lib/logs.js
@@ -4,6 +4,7 @@ var stream = require('stream');
 var split = require('binary-split');
 
 module.exports.log = log;
+module.exports.logStream = logStream;
 module.exports.fetch = fetch;
 
 /**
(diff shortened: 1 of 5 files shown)


lib/logs.js
```diff
@@ -22,6 +23,17 @@ function log() {
   logger.info.apply(null, arguments);
 }
 
+function logStream() {
+  var writable = new stream.Writable();
+
+  writable._write = function(line, enc, callback) {
+    log(line.toString());
```
rclark: Wouldn't the point of doing this be to get the lines passed through fastlog and prefixed?

PR #46, opened 2016-09-07
[wip] Smarter startedBy

### LINKED ISSUE #45: Better `startedBy` attribute on RunTask
This currently worker tasks just say `watchbot`. Other alternatives could be:
- the name of the worker library (bad)
- the CFN stack's name (better)
- the id of the watcher ecs:Service (overkill?)

Any of these would make it a bit easier to use the ECS API to get to your worker containers, for example to learn about their metrics refs #43 

cc @jakepruitt @yhahn

rclark:
As a first pass, tries to set the worker `startedBy` attributes to the watcher's service id.

- **potential benefit**: workers magically included in watcher service cloudwatch metrics
- **potential drawback**: workers counted towards watcher service desired capacity

Fixes #45


rclark:
Had to back away from trying to use the service id. There's really no way for the watcher container to know what service it is a part of without:

- knowing its own container name, can be passed as env through CFN template
- listTasks for everything running on the cluster, describeTasks for each
- find the task that has my container name
- use the same `startedBy` value

It just... doesn't seem worth it compared to just using the stack's name as `startedBy`, which is trivial.

PR #51, opened 2016-09-15
watchbot.progress returns promises, include worker script examples

rclark:
Like it says in the title. Depending on your promise preferences you may want this feature @l-r @rodowi? If so let me know and we can do a minor version tag.


l-r:
I personally prefer this to callbacks, but @rodowi is gonna be the one writing the code :)

rodowi:
👍  promises FTW

PR #54, opened 2016-09-19
Fix ref to WorkerRole for ProgressTablePermission

rodowi:
This will fix `WatchbotProgressTablePermission: The role with name WatchbotWorkerRole cannot be found` when creating a stack with reduce mode enabled.


rclark:
"instead of from a **fork**" I meant to say

rodowi:
k, trying again a new branch.

PR #55, opened 2016-09-20
Fix WorkerRole ref for ProgressTablePermission

rodowi:
Direct PR to apply patch #54


rodowi:
🚀  approved by inspector @rclark

PR #58, opened 2016-10-09
Add Metadata.EcsWatchbotVersion

yhahn:
For use with forthcoming tools to scan existing stacks and determine `ecs-watchbot` version in use.

cc @rclark @emilymcafee

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+### 0.0.13
+
+- Adds EcsWatchbotVersion to template Metadata
+
 ### 0.0.12
 
 - allow `workers` and `backoff` to be a ref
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -30,5 +31,9 @@ module.exports = {
    * @throws errors when there is overlap in logical resource names between
    * templates
    */
-  merge: cf.merge
+  merge: function() {
+    var template = cf.merge.apply(cf, arguments);
```
yhahn: Hrm, `cloudfriend.merge()` will throw though if you include multiple bots in a single stack ya? (https://github.com/mapbox/cloudfriend/blob/fae27a3b45b65b1f8d8863c4692e889176dc89b2/lib/merge.js#L30)

rclark: Yes, intentionally, to prevent any ambiguity where you were unaware that a property you set in one object overwrote another.

Right now the cloudfriend.merge checks are too strict -- they'll fail if you use the same logical name of a resource and output name, for example. I've been meaning to change this, and if that's what was stopping you I can do it right now.

PR #60, opened 2016-10-11
cloudformation-kms

yhahn:
Let all ecs-watchbot stacks do encryption/decryption of CF params. Requires prerequisite [cloudformation-kms](https://github.com/mapbox/cloudformation-kms) to be deployed.

- [x] integration test
- [x] review

cc @rclark @emilymcafee @aarthykc

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -260,6 +260,11 @@ module.exports = function(options) {
                 Effect: 'Allow',
                 Action: ['logs:CreateLogStream', 'logs:PutLogEvents', 'logs:FilterLogEvents'],
                 Resource: '*'
+              },
+              {
+                Effect: 'Allow',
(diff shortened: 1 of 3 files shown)


yhahn:
Currently `watchbot` is essentially a `devDependency` and not required at runtime by the worker container itself, right?

I think it would be good to keep it that way for now, i.e. `decrypt-kms-env` is best defined as an explicit dependency of each app, esp since there's also a python/awscli centric implementation as well (https://github.com/mapbox/decrypt-kms-env#v1x).

rclark:
You're right there's no reason to think that watchbot would be automatically part of every worker's dockerfile / image. Maybe it is just worth mentioning in the readme that while this'll encrypt `[secure]` parameters for you, you're responsible for decrypting them in your worker.

PR #63, opened 2016-10-15
Also retry on CPU constraint errors

yhahn:
Ran into watchbot stopping dead in its tracks on a job if it hit a RESOURCE:CPU error. Adds retry for this case as well.


yhahn:
Staged :+1:

Issue #17, opened 2016-05-20
Resource polling issues during deploys


rclark:
```
Error polling cluster resources: Container instance cannot be empty.
```

This seems to happen once as soon as the watcher starts up.

rclark:
I've seen this happening now at other times than just at watcher startup.

emilymdubois:
@rclark I can only recreate the watcher startup scenario, but it looks like this error is thrown when the [instances array in the `resources.available`](https://github.com/mapbox/ecs-watchbot/blob/de041ebd8d91230fa51cde194ffaf27188580175/lib/resources.js#L46) is empty, which causes the [`describeContainerInstances` function](https://github.com/mapbox/ecs-watchbot/blob/de041ebd8d91230fa51cde194ffaf27188580175/lib/resources.js#L48) to return a native AWS error.

A few questions here:

1. Is it expected that we don't have any registered instances during watcher startup?
2. If so, perhaps we can return a custom message in the event of ^^ and bypass [the describeContainerInstances call](https://github.com/mapbox/ecs-watchbot/blob/41ef6c9d4210ad7e7df0f9587ef31508de42c57f/lib/resources.js#L46-L68)?
3. Is this a scenario that we would want to handle differently during watcher startup vs. other times (i.e., maybe this is expected during startup, but not expected at other times)?

rclark:
> Is it expected that we don't have any registered instances during watcher startup?

Not really... because the watcher container that is making a request runs on a cluster that has to have container instances -- otherwise the watcher wouldn't have a place to run. However we do see it consistently on startup, so something systematic is at play here.

Is it possible that there's a container/permissions race-condition here? That the container launches before it has permission to ListContainerInstances?

> perhaps we can return a custom message in the event of ^^ and bypass the describeContainerInstances call?

Yeah we definitely can but... 

> Is this a scenario that we would want to handle differently during watcher startup vs. other times 

Yes. If this were happening consistently at times other than startup it would be a) confusing and b) probably indicate some AWS issue with the ECS API.

emilymdubois:
Closed via https://github.com/mapbox/ecs-watchbot/pull/67#issuecomment-256182416

PR #67, opened 2016-10-25
Resource polling issues

emilymdubois:
Refs https://github.com/mapbox/ecs-watchbot/issues/17

Here's my first pass at addressing the resource polling issues we've been seeing. I believe this was happening because [`update` clears out the status.instances array](https://github.com/mapbox/ecs-watchbot/blob/c24e9a6d9363c58138ef0ec314de5b95cbd0b5f9/lib/resources.js#L27-L33), which resources.available [consumes](https://github.com/mapbox/ecs-watchbot/blob/c24e9a6d9363c58138ef0ec314de5b95cbd0b5f9/lib/resources.js#L46) before the [update function runs](https://github.com/mapbox/ecs-watchbot/blob/c24e9a6d9363c58138ef0ec314de5b95cbd0b5f9/lib/resources.js#L31).

This PR places an interval in the resources.available function that checks the status.instances array every second up to 5 times, after which it returns the callback with an error message if the array never becomes populated.

@rclark, would love your thoughts on this 🙏

### DIFF
--- a/lib/resources.js
+++ b/lib/resources.js
@@ -51,9 +51,9 @@ module.exports = function(cluster, taskDef) {
         wrapper();
       } else if (status.instances.length === 0 && attempts === 5) {
         clearInterval(poller);
-        callback(new Error('No instances available after 5 polling attempts 3 seconds apart'));
+        callback(new Error('No instances available after 5 polling attempts 1 second apart'));
       }
-    }, 3000);
+    }, 1000);
 
     function wrapper() {
       ecs.describeContainerInstances({ containerInstances: status.instances, cluster: cluster }, function(err, data) {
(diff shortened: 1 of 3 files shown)


emilymdubois:
@rclark, swapped out the interval for an event emission, and added an error emission when there are no instances in the cluster. Tested out this branch by launching a stack and everything looks 👌

rclark:
Awesome! Please merge any time, but wait to tag this -- I think @mapsam is hot on your heels with another bugfix.

emilymdubois:
Thanks @rclark! @mapsam, will leave the tagging to you when you merge. Formatted my commit message so you should be able to plop it into the changelog.

PR #68, opened 2016-10-25
NotifyAfterRetry

mapsam:
Adds the option & logic for #2 

TODO

- [x] write a test! Since this is an option set through environment vars in the initial creation of a watchbot template, how can I mock the message logic retrying?

cc @rclark @emilymdubois


emilymdubois:
This is awesome @mapsam 🙇  publisher also needs this concept, so this is super helpful. Let me know if you need extra eyes with test writing.

emilymdubois:
@mapsam, here's how I would think about approaching tests for this addition:

**template.test.js**

To confirm that [this line is handling as expected](https://github.com/mapbox/ecs-watchbot/blob/82c4cc924d988bbf373d62d28d44764bb0291ed0/lib/template.js#L395), you could add some assertions about what this value looks like in the resultant `watch.Resources` object for the default and user specification scenarios.

**task.test.js**

There's already a test that [loops through each of the task outcomes](https://github.com/mapbox/ecs-watchbot/blob/82c4cc924d988bbf373d62d28d44764bb0291ed0/test/tasks.test.js#L188-L253). I think you would be fine adding a few more `envs` [here](https://github.com/mapbox/ecs-watchbot/blob/82c4cc924d988bbf373d62d28d44764bb0291ed0/test/tasks.test.js#L194-L203), and testing to ensure that you see the expected outcomes [here](https://github.com/mapbox/ecs-watchbot/blob/82c4cc924d988bbf373d62d28d44764bb0291ed0/test/tasks.test.js#L239-L245). The environments that you add would need to have a few extra parameters to test your message count handling. For example:

```js
{ exit: '4', MessageId: 'exit-4', ApproximateReceiveCount: 3, NotifyAfterRetries: 3 }
{ exit: '4', MessageId: 'exit-4', ApproximateReceiveCount: 4, NotifyAfterRetries: 3 }
```

lib/tasks.js
```diff
@@ -144,15 +153,13 @@ module.exports = function(cluster, taskDefinition, containerName, concurrency, s
...
-              return env;
-            }, {}),
+            env: taskEnvironment,
             outcome: task.containers.reduce(function(outcome, container) {
               var containerOutcome = (function() {
                 if (container.exitCode === 0) return tasks.outcome.success;
                 if (container.exitCode === 3) return tasks.outcome.fail;
-                if (container.exitCode === 4) return tasks.outcome.noop;
+                if (container.exitCode === 4 && 
+                    +taskEnvironment.ApproximateReceiveCount <= +taskEnvironment.NotifyAfterRetries) return tasks.outcome.noop;
```
mapsam: Maybe the default should be set to `1`, so every exit `4` will return `immediate` instead of `retry & notify`.

mapsam:
Huh, noticing that `taskEnvironment` doesn't contain any of the default values for `NotifyAfterRetries` ... maybe I'm looking at the wrong environment here. Continuing to dig.

lib/tasks.js
mapsam: `taskEnvironment.ApproximateReceiveCount` is always undefined here, same with `taskEnvironment.NotifyAfterRetries` if it's not explicitly set. Seems like there is a missed connection somewhere.

rclark: "is always undefined" in tests? The tests will need to explicitly set these when the tasks are run.

Also I don't think that you want this behavior to invoke on `container.exitCode === 4` -- you want this behavior whenever the script was about to `tasks.outcome.retry`, i.e., everything that's not exit code 0, 3, or 4.

mapsam: Oh, good to know - adding as a new `if` statement for anything not `0`, `3`, or `4` should be good. 

Didn't realize the `env` was a total mock in the task tests, just added new properties to them which is looking good.

mapsam: Fixed. There is now a check after `0`, `3`, and `4` to retry before notifying.

mapsam:
@rclark @emilymdubois i think this is ready!

mapsam:
```

Issue #62, opened 2016-10-12
describeContainerInstances limited to 100 at a time


rclark:
The request to [describeContainerInstances here](https://github.com/mapbox/ecs-watchbot/blob/5f63c78e91b2863af00ab6e44e05e73fbeb6c671/lib/resources.js#L46) will fail on a cluster with more than 100 instances.

emilymdubois:
Closed via https://github.com/mapbox/ecs-watchbot/pull/73

PR #73, opened 2016-10-27
Adds pagination for describeContainerInstances

emilymdubois:
Refs https://github.com/mapbox/ecs-watchbot/issues/62

This PR adds pagination for describeContainerInstances by [piggy-backing off the pagination already occurring for listContainerInstances](https://github.com/mapbox/ecs-watchbot/blob/93a3c4b59b51d87beb1baeb06b4cc572289a036e/lib/resources.js#L19-L28).

The test that's causing the 🏄  of errors is the [resource polling error test](https://github.com/mapbox/ecs-watchbot/blob/93a3c4b59b51d87beb1baeb06b4cc572289a036e/test/main.test.js#L312-L322). I can't quite figure out what's preventing the [finish event](https://github.com/mapbox/ecs-watchbot/blob/93a3c4b59b51d87beb1baeb06b4cc572289a036e/test/main.test.js#L316) from firing, since it looks like the [end function](https://github.com/mapbox/ecs-watchbot/blob/93a3c4b59b51d87beb1baeb06b4cc572289a036e/lib/main.js#L202-L204) is being called. Nothing jumps out to me immediately, but perhaps some seasoned eyes can spot something I'm missing.

cc @rclark

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -62,7 +62,7 @@ module.exports = function(config) {
       resources.status.available.cpu, resources.status.registered.cpu,
       resources.status.available.memory, resources.status.registered.memory
     );
-
+    
     timer = setTimeout(status, 30000).unref();
   }
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -62,7 +62,7 @@ module.exports = function(config) {
       resources.status.available.cpu, resources.status.registered.cpu,
       resources.status.available.memory, resources.status.registered.memory
     );
-    
+
     timer = setTimeout(status, 30000).unref();
   }
 
--- a/lib/resources.js
+++ b/lib/resources.js
@@ -24,7 +24,7 @@ module.exports = function(cluster, taskDef) {
         status.availableInstances = status.availableInstances.concat(available);
         if (data.nextToken) return listInstances(data.nextToken, callback);
         callback();
-      })
+      });
     });
   }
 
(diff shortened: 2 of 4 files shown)


rclark:
I think the problem is that in this test case the `HasInstances` event never emits. This prevents the main watchbot loop from cycling. `main.end()` does get called, but then another iteration through the loop never starts. Once one does, it'll pick up that it is supposed to shut down and emit the `finish`.

emilymdubois:
That makes sense, thanks @rclark! I restructured the test so that it [makes its assertion about log contents after 1.8s](https://github.com/mapbox/ecs-watchbot/blob/080f7dd0428df03a9c10f79f379b43e6e9546118/test/main.test.js#L317-L322) instead of waiting for the `finish` event.

Do you think [this is an appropriate order of operations](https://github.com/mapbox/ecs-watchbot/blob/080f7dd0428df03a9c10f79f379b43e6e9546118/lib/resources.js#L41-L43) (i.e., that the `HasInstances` never fires and halts the main loop if there's an ECS error either with listing or describing container instances)?

rclark:
The loop is going to be blocked until the `HasInstances` event fires **and** there are instances in the `status.instances` array. This is fine. The scenario that comes up in the test here is totally bizzaro / not real-life like we talked about yesterday.

emilymdubois:
👍 Sounds good @rclark! Let me know if you have any reservations about me merging this. Otherwise, I'll merge, wait for https://github.com/mapbox/ecs-watchbot/pull/74, and roll a new release.

cc @mapsam

mapsam:
#74 merged - apologies for the cruft!

jacquestardie:
Just checking in, this hasn't been made it into a release yet, correct? If it hasn't, I'm happy to grab!

rclark:
@jacquestardie -- @emilymdubois is also trying to wrap #75 into the next ecs-watchbot release.

PR #83, opened 2016-12-06
add privileged parameter to TaskDefinition

mapsam:
This adds a watchbot parameter `privileged` that provides an opportunity to run the container with elevated credentials.

cc @emilymdubois @rclark @springmeyer

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -369,7 +369,6 @@ module.exports = function(options) {
           Memory: options.reservation.memory,
           Cpu: options.reservation.cpu,
           Privileged: options.privileged,
-          ReadonlyRootFilesystem: options.privileged ? false : true,
           Environment: unpackEnv(prefixed, options.env),
           MountPoints: mounts.mountPoints,
           Command: options.command,
--- a/lib/template.js
+++ b/lib/template.js
@@ -137,7 +137,8 @@ module.exports = function(options) {
     logGroup: cf.ref(prefixed('LogGroup')),
     topic: cf.ref(prefixed('Topic')),
     queueUrl: cf.ref(prefixed('Queue')),
-    queueArn: cf.getAtt(prefixed('Queue'), 'Arn')
+    queueArn: cf.getAtt(prefixed('Queue'), 'Arn'),
+    queueName: cf.getAtt(prefixed('Queue'), 'QueueName')
   };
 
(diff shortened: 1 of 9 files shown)


rclark:
Do you also need to consider ReadOnlyRootFileSystem on the task definition? https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-readonlyrootfilesystem?

mapsam:
Good point @rclark. @springmeyer brings up another point that we may need to edit `ulimit` info as well: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html#cfn-ecs-taskdefinition-containerdefinition-ulimits - if the goal is to edit core files.

mapsam:
Confirmed this works - ready for a final review @rclark @emilymdubois!

lib/template.js
```diff
@@ -364,6 +367,8 @@ module.exports = function(options) {
           Image: cf.join([cf.accountId, '.dkr.ecr.', cf.findInMap('EcrRegion', cf.region, 'Region'), '.amazonaws.com/', options.service, ':', options.serviceVersion]),
           Memory: options.reservation.memory,
           Cpu: options.reservation.cpu,
+          Privileged: options.privileged,
+          ReadonlyRootFilesystem: options.privileged ? false : true,
```
rclark: Don't actually need the `? false : true` here, right?

mapsam: I wasn't sure, so I played it safe. Think it should just be `false` all the time?

rclark: I mean that `options.privileged` is already either `true` or `false`. Do you need another conditional?

mapsam: Ah, I see! Yeah, this can just be `!options.privileged` if it's necessary.

mapsam:
@rclark confirmed this works just fine without the ReadOnlyRootFilesystem addition.

rclark:
Thanks @mapsam, looks great.

mapsam:
Well, apparently `git rebase master` and `git push origin <branch>` isn't the right way to do things from your local checkout. I squashed the duplicate history on the merge, so we should be fine. Sorry for the cruft!

PR #93, opened 2017-01-14
notifyAfterRetries back to the watcher

mapsam:
Third time's a charm. This moves the notifyAfterRetries parameter back into the watcher to be handled within main.js instead of the workers.

* removes leftover task tests
* adds new logic to main.js that munges the `taskDefinition.outcome` based on the parameter

Need to add tests into tasks.test.js still - might need some spirit guidance from @rclark though.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -102,6 +102,10 @@ module.exports = function(config) {
 
       // Handle each finished task
       status.forEach(function(finishedTask) {
+
+        // retry before notifying, if it's been set in options.notifyAfterRetries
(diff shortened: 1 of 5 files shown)


lib/main.js
```diff
@@ -102,6 +102,10 @@ module.exports = function(config) {
 
       // Handle each finished task
       status.forEach(function(finishedTask) {
+
+        // retry before notifying, if it's been set in options.notifyAfterRetries
+        if (finishedTask.env.ApproximateReceiveCount <= process.env.NotifyAfterRetries) finishedTask.outcome = tasks.outcome.noop;
```
rclark: I believe you may still need to make sure these are coerced from strings --> numbers

Issue #97, opened 2017-02-01
Allow AlarmTopic or AlarmEmail


emilymcafee:
We should support publishing failure notifications to an SNS topic directly, instead of creating the topic/email subscription by default. If AlarmTopic is provided instead of AlarmEmail, then no-op on creating a separate topic, just publish directly to the one provided.

cc @mapbox/platform

emilymcafee:
I guess watchbot lingo would be notificationEmail => notificationTopic..

PR #98, opened 2017-02-01
Optionally provide notificationTopic instead of notificationEmail

emilymcafee:
Fixes https://github.com/mapbox/ecs-watchbot/issues/97

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -29,6 +29,8 @@ var table = require('@mapbox/watchbot-progress').table;
  * container that will be deployed. Defaults to the installed version
  * @param {string|ref} options.notificationEmail - an email address to receive
  * notifications when processing fails.
+ * @param {string|ref} options.notificationTopic - an ARN of the SNS topic to receive
+ * notifications when processing fails.
  * @param {string} [options.prefix='Watchbot'] - a prefix that will be applied
  * to the logical names of all the resources Watchbot creates. If you're
  * building a template that includes more than one Watchbot system, you'll need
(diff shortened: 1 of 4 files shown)


lib/template.js
```diff
@@ -29,6 +29,8 @@ var table = require('@mapbox/watchbot-progress').table;
  * container that will be deployed. Defaults to the installed version
  * @param {string|ref} options.notificationEmail - an email address to receive
  * notifications when processing fails.
+ * @param {string|ref} options.notificationTopic - an ARN of the SNS topic to receive
```
rclark: `[]` around the `options.notificationTopic` and `options.notificationEmail` to indicate that it they're optional

lib/template.js
```diff
@@ -154,7 +156,10 @@ module.exports = function(options) {
   if (options.reduce) reduce(prefixed, resources, options, references);
   var mounts = mount(options.mounts);
 
-  resources[prefixed('NotificationTopic')] = {
+  if (options.notificationTopic && options.notificationEmail) throw new Error('Cannot provide both notificationTopic and notificationEmail.');
```
rclark: I think you could allow both to be specified, and should assert that at least one is given. In the event that both exist, you'd just have 2 AlarmActions in the array

emilymcafee: In the event that both exist, it's pretty straightforward to have 2 AlarmActions, but we'd need to do some re-writing in https://github.com/mapbox/ecs-watchbot/blob/master/lib/notifications.js to accept an array of topics, yeah?

rclark: Ok, if you don't think that's worthwhile then 🚢

emilymcafee:
I went down the road of implementing multiple notification topics, but we need this to be a string so that [it can be passed into the worker container's environment](https://github.com/mapbox/ecs-watchbot/blob/39e743f5742e0bd956e6bb1fbff229cf491aa02c/lib/template.js#L421). Punting and requiring one or the other in this iteration.

PR #101, opened 2017-02-04
Remove resources, switch to just brute-forcing to run tasks

### LINKED ISSUE #79: Remove cluster resource availability polling
[This section of the main loop](https://github.com/mapbox/ecs-watchbot/blob/bac01e7e261d0159b4ec580482c0e8e034e6599b/lib/main.js#L121-L133) checks the cluster's available resources in an attempt to avoid trying to launch a worker task when the cluster does not have sufficient resources. The primary motivation for adding this is that it helps avoid alarm-spam if the cluster is out of resources but messages need to be processed. Each attempt to launch a container that failed would lead to an alarm.

I think this concept needs to be revisited for a few reasons:
- #76 could provide a better alternative to the 1:1 alarms that motivated this in the first place
- An ECS cluster might want to keep track of the number of tasks that wanted to be placed but couldn't find sufficient resources. This could be used as a metric by which to add scale up the cluster's capacity.
- Doing this polling adds [almost 100 lines of code to test and maintain](https://github.com/mapbox/ecs-watchbot/blob/bac01e7e261d0159b4ec580482c0e8e034e6599b/lib/resources.js), and provides a pretty vague benefit.
- The polling results in the need to routinely list all the EC2s in the cluster. We could avoid making these API calls and it would have some impact on overall API request rate throttling that we've seen under some situations.

jakepruitt:
Drops all resource polling, and trusts that the retry logic in `task.runTask` will handle the filled-cluster-resources situation. This hopefully will resolve some of the improper watcher deaths and watcher hangs we've seen recently.

The only indication of failed task placement is an extra error log message, which could someday be turned into a cloudwatch-log-based metric.

Closes #79 

cc/ @rclark @yhahn


lib/main.js
```diff
@@ -44,36 +46,22 @@ module.exports = function(config) {
...
-      resources.status.available.cpu, resources.status.registered.cpu,
-      resources.status.available.memory, resources.status.registered.memory
+      Object.keys(messages.inFlight).length
     );
 
-    timer = setTimeout(status, 30000).unref();
-  }
-
-  resources.once('update', status);
+    setTimeout(status, 30000).unref();
```
rclark: If you increase this to 60000 and make it `log.info`, that might be another useful watcher log that could be metricified.

lib/template.js
```diff
@@ -229,6 +229,32 @@ module.exports = function(options) {
     }
   };
 
+  resources[prefixed('FailedTaskPlacementMetric')] = {
+    Type: 'AWS::Logs::MetricFilter',
+    Properties: {
+      FilterPattern: 'task did not run: RESOURCE:',
+      LogGroupName: cf.ref(prefixed('LogGroup')),
+      MetricTransformations: [{
+        MetricName: cf.stackName + '-failed-task-placement',
```
rclark: How does this work? If you expand the Cloudfriend reference, you get

```
"MetricName": { "Ref": "AWS::StackName" } + "-failed-task-placement"
```

I'm swinging towards keeping the collection here in filters. But let's rename them to

```
cf.sub('FailedTaskPlacement-${AWS::StackName})
cf.sub('WorkerErrors-${AWS::StackName})
```

Issue #100, opened 2017-02-03
Support ephemeral volumes


rclark:
Currently `watchbot.template()` requires you provide `mounts` as strings indicating the container and host file system paths. This precludes you from being able to specify a volume where host disk space is ephemeral / managed by docker.

The caller should be able to specify 

```
"mounts": "/path/inside/container"
```

... and this should build a task definition that'll use an ephemeral volume.

rclark:
Closed in #102

PR #102, opened 2017-02-06
Adds ephemeral mount volume compatibility

emilymdubois:
Refs https://github.com/mapbox/ecs-watchbot/issues/100

I believe swapping the source volume name for `database_scratch` is the first step in supporting ephemeral volumes. `database_scratch` will be used when the mount pathway begins with `/mnt/tmp`. @rclark, is this on the right track?


lib/template.js
```diff
@@ -714,7 +714,7 @@ function mount(mountStrs) {
 
     var host = mountStr.split(':')[0];
     var container = mountStr.split(':')[1];
-    var name = 'mnt-' + i;
+    var name = /^\/mnt\/tmp/.test(host) ? 'database_scratch' : 'mnt-' + i;
```
rclark: I think you want to see if `mountStr` contains a `:`. If it does, then this is a persistent mount (like we already support). If it does not, then this is an ephemeral mount which only specifies a path on the container. What you name the mount is less important -- it just has to be consistent between the two parts of the TaskDefinition.

emilymdubois:
Thanks for the feedback @rclark :bowing_woman: I think this makes more sense now. Moved things around so if only a filepath is specified (e.g., `.Mounts: '/mnt/tmp'`), the created resources look like:

```js
"containerDefinitions": [
  {
    "mountPoints": [
      {
        "sourceVolume": "mnt-0",
        "containerPath": "/mnt/tmp/"
      }
    ]
  }
],
"volumes": [
  {
    "name": "mnt-0",
    "host": {}
  }
]
```

lib/template.js
```diff
@@ -712,11 +712,13 @@ function mount(mountStrs) {
   mountStrs.split(',').forEach(function(mountStr, i) {
     if (!mountStr.length) return;
 
-    var host = mountStr.split(':')[0];
-    var container = mountStr.split(':')[1];
     var name = 'mnt-' + i;
+    var persistent = /:/.test(mountStr);
+    var host = persistent ? { SourcePath: mountStr.split(':')[0] } : {};
```
rclark: Definitely should double-check this in a staging scenario. I'm pretty sure it works if there's just no `.Host` property, but perhaps `.Host: {}` works too.

emilymdubois:
Tested this out and all looks well with specifying an ephemeral mount:

- CloudFormation: `mounts: '/mnt/tmp/emdb'`
- Worker code:

```js
watchbot.log(require('fs').existsSync('/mnt/tmp/emdb')); // logged `true`
watchbot.log(require('fs').readdirSync('/mnt/tmp/emdb')); // logged `[]`
require('fs').writeFileSync('/mnt/tmp/emdb/test.txt', 'test string');
watchbot.log(require('fs').readdirSync('/mnt/tmp/emdb')); // logged `test.txt`
```

@rclark, any other updates or gotchas you spot?

test/template.test.js
```diff
@@ -125,6 +126,9 @@ test('[template] webhooks but no key, no references', function(assert) {
   assert.ok(watch.Resources.testService, 'service');
   assert.notOk(watch.Resources.testProgressTable, 'progress table');
   assert.notOk(watch.Resources.testProgressTablePermission, 'progress table permission');
+  assert.ok(_.findWhere(watch.Resources.testWorker.Properties.ContainerDefinitions[0].MountPoints, { ContainerPath: '/mnt/tmp', SourceVolume: 'mnt-2' }), 'ephemeral mount point check');
```
rclark: Let's do this without underscore. No reason really other than minimizing package dependencies. Have you used [array.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)?

emilymdubois: @rclark I have not, but seems like it's the native JS equivalent of what I was trying to do. Swapping in!

emilymdubois:
@rclark added a note under the 1.5.0 header. Merging but not holding off on version bumping, tagging, and publishing.

emilymdubois:
Just kidding, we're on version 1.5.0 (record is missing from the changefile). Moved to 1.6.0.

Issue #108, opened 2017-02-08
Use container exit reason instead of task exit reason


rclark:
In alarm notifications, we look at a JSON object representing a task in order to interpret how / why the task exited. One thing we do is convert that into a log that comes from the watcher. It looks something like this:

```
[Mon, 06 Feb 2017 06:33:04 GMT] [watchbot] [c12d6a26-e2e3-4684-a28b-4801dd6ed6a2] finished | outcome: success | reason: Essential container in task exited | duration: 2s
```

We also create email notifications if a task exits non-zero. The email interprets the same object to print something like:

```
At Tue, 07 Feb 2017 23:48:26 GMT, processing message 5781849d-9083-46e2-918e-cba45faf7783 failed on blah-blah-blah

Task outcome: return & notify

Task stopped reason: Essential container in task exited

Message information:
MessageId: 5781849d-9083-46e2-918e-cba45faf7783
...
```

In both cases, the **reason** is pretty useless. This is because we're parsing out [the task's stoppedReason](https://github.com/mapbox/ecs-watchbot/blob/cwe/lib/tasks.js#L164), which is that, almost without fail, an "essential container in task exited".

Instead, we should be printing out **the container's stoppedReason**, which is almost always more interesting, like:

```
CannotPullContainerError: failed to register layer: devicemapper: Error running deviceResume dm_task_run failed
```

We'd pull this from `task.containers[0].reason`. **Warning**: sometimes this will be empty. I think it might be empty in the event that the task exited `0`.

cc @emilymdubois

emilymdubois:
@rclark, curious if there may be argument for returning `task.StoppedReason` and perhaps `task.containers[0].exitCode` in the event that `task.containers[0].exitCode` does not equal 0? I'm seeing some examples like the following, in which the task did not exit `0`, but also did not return a `task.containers[0].reason`:

```js
{
  tasks: [
    {
      containers: [
        lastStatus: 'STOPPED',
        exitCode: 137
      ]
    }
  ],
  stoppedReason: 'Scaling activity initiated by (deployment <deployment-id>)'
}
```

In the event that ☝️ happens, it may be helpful to have the `reason` provide something like `Scaling activity initiated by (deployment <deployment-id>), container exit code 137`.

What are your thoughts @rclark?

rclark:
[...]

PR #109, opened 2017-02-08
Use container exit reason when available

emilymdubois:
Refs https://github.com/mapbox/ecs-watchbot/issues/108

This PR adds some handling around the `reason` provided when a task finishes:

* If all of the containers have `task.containers[n].exitCode: 0`, return the string `success`.
* If any of the containers have `task.containers[n].exitCode: !0` and a `task.containers[n].reason`, return the container `task.containers[n].reason`. In the event that multiple containers exit with non-0, return the first `task.containers[n].reason` encountered.
* If any of the containers have `task.containers[n].exitCode: !0` and do not have a `task.containers[n].reason`, return `task.stoppedReason`.
* Updates changelog, assuming this will roll out with a few other updates in version 1.5.0

@rclark any gotchas and/or recommendations for staging?

### DIFF
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -132,12 +132,26 @@ module.exports = function(cluster, taskDefinition, containerName, concurrency, s
         }).map(function(task) {
 
           /**
-           * An task's environment used as a reference for the {@link finishedTask}
+           * A task's environment used as a reference for the {@link finishedTask}
            */
           var taskEnvironment = task.overrides.containerOverrides[0].environment.reduce(function(env, item) {
             env[item.name] = item.value;
             return env;
           }, {});
+
+          /**
+           * A task's reason for finishing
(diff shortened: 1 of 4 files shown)


lib/tasks.js
```diff
@@ -135,14 +135,28 @@ module.exports = function(cluster, taskDefinition, containerName, concurrency, s
...
+           * stopped container reason is provided return that. Otherwise, return
+           * stopped task reason.
+           */
+          var success = task.containers.every(function(c) { return c.exitCode === 0; });
+          var containerReason = task.containers.find(function(c) { return c.reason; });
+
+          var taskReason;
+          if (task.stoppedReason) taskReason = task.stoppedReason;
+          if (containerReason) taskReason = containerReason.reason;
+          if (success) taskReason = 'success';
```
rclark: :+1: beautiful

Issue #105, opened 2017-02-06
Create script for checking available capacity for new workers on a given cluster


jakepruitt:
From #103:

> Write a function to determine exactly how many tasks could currently be placed on a cluster. This has to look at each EC2, and determine how many tasks could be placed on it, then move on to the next ec2. As an npm run kind of script, this can help watchbot users answer their own questions about concurrency.

I imagine this script would live in the ecs-watchbot repo and take a parameter for the stackname of the ecs-watchbot stack and the stackname of the cluster.

cc/ @rclark @emilymdubois

rclark:
I think the cluster's ARN could be deduced from the stack's template.

jakepruitt:
@rclark so you would need to run this from the folder of the ecs-watchbot stack you're interested in?

rclark:
That or make a CloudFormation [GetTemplate API call](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudFormation.html#getTemplate-property).

emilymdubois:
Closed by https://github.com/mapbox/ecs-watchbot/pull/110

PR #110, opened 2017-02-08
Script to check for worker capacity on cluster

emilymdubois:
Refs https://github.com/mapbox/ecs-watchbot/issues/105

Initializing a PR for the worker capacity on cluster script.

- [x] Get cluster ARN given stack's name and region
- [x] Query worker capacity for each of the EC2s

### DIFF
--- a/bin/worker-capacity.js
+++ b/bin/worker-capacity.js
@@ -4,6 +4,6 @@ var argv = require('minimist')(process.argv.slice(2));
 var run = require('../lib/capacity').run;
 
 run(argv, (err, result) => {
-  if (err) console.log('\n' + err + '\n');
-  if (result.capacity) console.log('\n' + result.cluster + ' currently has enough space for an additional ' + result.capacity + ' ' + argv.stack + ' workers.\n');
+  if (err) console.log('\n%s\n', err);
+  if (result.capacity) console.log('\n%s currently has enough space for an additional %s %s workers.\n', result.cluster, result.capacity, argv.stack);
 });
--- a/bin/worker-capacity.js
+++ b/bin/worker-capacity.js
@@ -5,5 +5,5 @@ var run = require('../lib/capacity').run;
 
 run(argv, (err, result) => {
   if (err) console.log('\n%s\n', err);
-  if (result.capacity) console.log('\n%s currently has enough space for an additional %s %s workers.\n', result.cluster, result.capacity, argv.stack);
+  if (result.capacity) console.log('\n%s currently has enough space for an additional %s %s workers.\n', result.cluster, result.capacity, result.stack);
 });
--- a/bin/worker-capacity.js
+++ b/bin/worker-capacity.js
@@ -1,91 +1,9 @@
 #!/usr/bin/env node
 
-var AWS = require('aws-sdk');
 var argv = require('minimist')(process.argv.slice(2));
+var run = require('../lib/capacity').run;
 
-module.exports = {};
-module.exports.run = run;
-module.exports.getClusterArn = getClusterArn;
-module.exports.getReservations = getReservations;
-module.exports.listInstances = listInstances;
-module.exports.calculateRoom = calculateRoom;
-
-// run(argv, (err, result) => {
-//   if (err) console.log('\n' + err + '\n');
-//   if (result.capacity) console.log('\n' + result.cluster + ' currently has enough space for an additional ' + result.capacity + ' ' + argv.stack + ' workers.\n');
-// });
-
-function run(argv, callback) {
-  getClusterArn(argv, (err, clusterArn) => {
-    if (err) return callback(err);
-    getReservations(argv, (err, rsvps) => {
-      if (err) return callback(err);
-      listInstances(clusterArn, (err, resources) => {
-        if (err) return callback(err);
-        var result = {
-          capacity: calculateRoom(resources, rsvps),
-          cluster: clusterArn.match(/arn:aws:ecs:.*:\d*:cluster\/(.*)/)[1]
-        };
-        return callback(null, result);
-      });
-    });
-  });
-}
-
-function getClusterArn(argv, callback) {
-  /* Confirm that region and stack are provided */
-  if (!argv.region || !argv.stack) {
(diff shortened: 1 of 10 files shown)


bin/worker-capacity.js
```diff
@@ -0,0 +1,29 @@
...
+  /* Confirm that region and stack are provided */
+  if (!argv.region || !argv.stack) {
+    var text = 'Usage:   worker-capacity --region <region>  --stack <stack_name>\n';
+    text += 'Example: worker-capacity --region us-east-1 --stack ecs-telephone-staging';
+    return callback(text);
+  }
+
+  /* Get the CloudFormation template */
+  var cloudformation = new AWS.CloudFormation({ region: argv.region });
+  cloudformation.describeStacks({ StackName: argv.stack }, function(err, res) {
```
rclark: This actually doesn't always work. Most of the time our stacks use a parameter, but you can make a watchbot stack that doesn't. 

Instead of `describeStacks` you should make a `getTemplate` API request. Use that to look up [the Cluster that the Watcher is configured to launch its service into](https://github.com/mapbox/ecs-watchbot/blob/9e56cefc7f23a23469855d860e7ec1f995b24e14/lib/template.js#L515)

bin/worker-capacity.js
```diff
@@ -0,0 +1,29 @@
...
+    if (err) return callback(new Error(err));
+    return callback(null, res.Stacks[0].Parameters.find(function(p) { return p.ParameterKey === 'Cluster' }).ParameterValue);
```
rclark: These array manipulation functions are fun places to use `=>` functions because the syntax is so brief:

```
.Parameters.find(function(p) { return p.ParameterKey === 'Cluster' })

could be

.Parameters.find((p) => p.ParameterKey === 'Cluster')
```

bin/worker-capacity.js
emilymdubois: @rclark yeah, I tried `getTemplate` and when I specified both `Original` and `Processed`, I never was able to access a cluster ARN, just a `{"Ref": "Cluster"}` 🤔

emilymdubois: An example of the kind of request I'm making:

```js
var AWS = require('aws-sdk');
var cloudformation = new AWS.CloudFormation({ region: argv.region });
cloudformation.getTemplate({
  StackName: argv.stack,
  TemplateStage: 'Processed'
}, function(err, res) {
  if (err) return callback(new Error(err));
  console.log(res.TemplateBody); // can't find cluster ARN, just references to it
});
```

rclark: oh yeah 🤔 

How about we add an output to `watchbot.template()` that we know will provide the ARN via describeStacks

emilymdubois: @rclark I pushed up a stop-gap that evaluates if `.WatchbotService.Properties.Cluster` is a string or object. If it's a string, return that; if it's an object, make a `cloudformation.describeStacks` request to get the parameter value for the reference variable. I think the risk, though, is that the cluster ARN may not be a stack parameter (e.g., it could be resolved through `.Mappings`), so I think the `.Outputs` method might be more reliable here 👍

jacquestardie:
this will be so nice!

emilymdubois:
@rclark & @jakepruitt, I've got an initial pass at a capacity script. Here's what it does:

1. Fetches the cluster ARN from the given stack's `.Outputs` property.
2. Fetches the worker memory and CPU reservations from a `cloudformation.getTemplate` API call.
3. For each instance in the cluster, fetches the memory and CPU `.remainingResources`.
4. Determines how many workers could fit on each EC2 based on memory and based on CPU availability and requirements. The lowest value of the two is taken per instance, and summed together.

Do you see any issues with this approach before I write out tests?

emilymdubois:
@rclark thanks for the feedback! Would you mind giving this a review?

One thing I noticed while testing is that a stack update alone didn't expose the cluster ARN via the `Outputs` property; I needed to delete and re-create my stack. Not sure if that's something worth mentioning in the readme, or is understood knowledge about CloudFormation.

lib/capacity.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function calculateRoom(resources, reservations) {
+  var taskCapacities = [];
+  for (var i in resources) {
+    var workerCpu = resources[i].find((e) => { return e.name === 'CPU'; }).integerValue / reservations.Cpu;
+    var workerMemory = resources[i].find((e) => { return e.name === 'MEMORY'; }).integerValue / reservations.Memory;
+    taskCapacities.push(Number(Math.min(workerCpu, workerMemory).toFixed(0)));
+  }
+  return taskCapacities.reduce((a, b) => { return a + b; });
```
rclark: If you want to take the reduce journey a step further, you can avoid having 2 loops by either:

- return Object.keys(resource).reduce(function(free, key) { ...
- have the listInstances function return an array instead of an object indexed by instance id

emilymdubois:
@rclark thanks for the review! Made some updates to make the `listInstances` and `calculateRoom` functions a little cleaner, and fixed a test bug.

Do you know of a way to make [this logic](https://github.com/mapbox/ecs-watchbot/blob/capacity-script/test/capacity.test.js#L37-L42) a little cleaner with sinon functions? I tried a few different ways to return different callbacks on the first and second stub calls, but couldn't quite get it right.

rclark:
I think the full-blown sinon-ified method would look like:

```js
var describe = AWS.stub('ECS', 'describeContainerInstances');
describe.onCall(0).yields(null, fixtures.describeContainerInstances0);
describe.onCall(1).yields(null, fixtures.describeContainerInstances1);
```

Does that work?

emilymdubois:
@rclark, ah that does work! I was attempting the following, without realizing that assigning the stub to a variable name means you can't string the functions together:

```js
var describeContainerInstances = AWS.stub('ECS', 'describeContainerInstances')
.onCall(0).yields(null, fixtures.describeContainerInstances0)
.onCall(1).yields(null, fixtures.describeContainerInstances1)
```

Anyway, I feel good about merging this! 👍

PR #111, opened 2017-02-09
2nd SQS for cloudwatch task-state-change events

### LINKED ISSUE #82: Utilize CloudWatch events to monitor task state changes
In the current iteration of watchbot, the watcher container's job is to perform these actions in an infinite loop:

Work SQS ♻️ polling watcher ➕ ECS ♻️ polling watcher

- poll the SQS queue to find work that needs doing
- launch a worker container to perform the work
- poll ECS `DescribeTasks` roughly every second until the task is finished
- interpret the final state of the worker container and deal with deleting or returning the SQS message.
- while operating below the desired worker concurrency, keep repeating this process.

This system works until you start running lots of watchbot systems in a single AWS account. You can begin to run into issues with `DescribeTasks` request rates. The introduction of [task state-change events in CloudWatch events](https://aws.amazon.com/about-aws/whats-new/2016/11/amazon-ec2-container-service-now-sends-state-changes-to-cloudwatch-events/) offers a new way for the watcher to function. I can think of a couple of ways to utilize this.

## Replace ECS polling with SQS polling

Work SQS ♻️ polling watcher ➕ CW Event ➡️ Task SQS ♻️ polling watcher

1. A watchbot stack consists of **two** SQS queues
  - the "work" queue like we already have
  - a "task status" queue where cloudwatch events are forwarded

2. The watchbot resources generated by `watchbot.template()` would generate a [CloudWatch events Rule](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html) which would forward task state-change events to the "task status" queue. Here is an example filter pattern for those events:

  ```json
  {
    "source": ["aws.ecs"],
    "detail-type": ["ECS Task State Change"],
    "detail": {
      "clusterArn": [
        "arn:aws:ecs:us-east-1:123456789012:cluster/ecs-cluster-processing-staging"
      ]
    }
  }
  ```

3. The watcher container would drop [all the code related to polling for task status via `DescribeTasks` API calls](https://github.com/mapbox/ecs-watchbot/blob/d16b00ed892f424848c6aaf8e0b830e77799172b/lib/tasks.js#L114-L181), instead replacing this with code that polls the "task status" queue, filtering out extraneous messages and interpretting relevant ones.

- **Pro**: dual-SQS polling offers some level of similarity to what watchbot already does
- **Con**: doesn't manage to get rid of any of the existing polling complexity

## Let SNS push notifications to the watcher

Work SQS ♻️ polling watcher ➕ CW event ➡️ SNS ➡️ ALB ➡️ watcher https service

1. The CloudFormation resources provided by `watchbot.template()` would add:
  - A new SNS topic to receive "task status" notifications
  - A CloudWatch Events Rule to forward task state-change events to the new SNS topic
  - A load balancer to expose an `https` endpoint
  - An SNS subscription that would push notifications to the load balancer

2. The watcher containers would sit behind the load balancer and run an `https` service that responds to the events that have been forwarded to the watcher containers. Similar to above, this would filter out extraneous messages and translate relevant ones into deleting or returning messages to the "work" queue.

- **Pro**: one less polling step
- **Con**: replace polling complexity with new systems & cost involved in exposing an https service for the watcher to listen for state-change events.

cc @mapbox/platform

rclark:
Adds a 2nd SQS queue. CloudWatch events for ECS task state changes are filtered by TaskDefinition and `lastStatus: STOPPED`. Matches are sent directly to this queue. This gives us a backlog of tasks that are finished that need to be taken care of.

`tasks.js` implements a continuous polling of that 2nd queue. As it finds messages there, it:
- filters them based on task ARN, matching against a list of tasks it has started
- ARNs that don't match are returned to the queue. These are messages from another watcher
- ARNs that do match are deleted from the queue, and stashed in memory.

When `main.js` calls `tasks.pol()`, it receives the set of finishedTasks that have been stashed in memory, and that cache is cleared.

cc @jakepruitt @emilymdubois 
fixes #82


rclark:
Requisite gif representing the watcher's dual-queue setup:

8o5gznc

Issue #90, opened 2017-01-11
Support cloudfriend.join/sub helpers for mount points


rodowi:
Both of these examples will fail:

```js
mounts: f.sub('/mnt/persist/${stack}:/tmp/foo', { stack: f.stackName }),

mounts: f.join(':', [
    f.join('/', ['mnt', 'persist', f.stackName]),
    f.join('/', ['mnt', 'persist', f.stackName])
  ])
```

due to
```
/.../node_modules/watchbot/lib/template.js:687
  mountStrs.split(',').forEach(function(mountStr, i) {
            ^

TypeError: mountStrs.split is not a function
```

Can we eval [cloudfriend](https://github.com/mapbox/cloudfriend) helpers when parsing mount points?
https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L685-L707

This could be a helpful feature to set dynamic mount points on shared volumes

rclark:
> Can we eval cloudfriend helpers when parsing mount points?

We really can't unfortunately. The evaluation of the javascript code occurs **before** CloudFormation variables are set and intrinsic functions evaluated.

There **may** be ways to interpret the CFN intrinsic functions that could help us, but the variety of ways you may end up specifying this is unbounded:

- as `"Fn::Sub"`
- as `"Fn::Join"`
- as `"Fn::Join"` with sub `"Fn::Join"`s
- as a simple "Ref" to a parameter value
- ... etc

The primary thing [the linked code needs to do](https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L685-L707) is break apart the string into mount points and volumes it can put into a TaskDefinition. We could probably define a custom JS object structure that the `mount` parameter could accept, where you'd break these apart yourself. Something like

```mounts: {
  mountPoints: ['/tmp/foo'],
  volumes: [
    { 'Fn::Sub': '/mnt/persist/${AWS::StackName}' }
  ]
}```

If the `mounts` function knows that it'll either receive a string or this kind of object, then it should be able to properly construct the TaskDefinition.

rodowi:
🤔  I see. So there's no way to evaluate CFN intrinsic functions at that point. Even though we break the mount definitions we'd still need to build or find a CFN parser.

rclark:
@emilymdubois this might be a nice followup to your work with mount points.

emilymdubois:
Closed by https://github.com/mapbox/ecs-watchbot/pull/115

PR #115, opened 2017-02-10
Support cloudfriend operators in mount points

emilymdubois:
Refs https://github.com/mapbox/ecs-watchbot/issues/90

This PR adds the option to also provide a mount point object:

```js
// two persistent mounts and one ephemeral
mounts: {
  container: ['/var/tmp', '/mnt/data', '/mnt/tmp'],
  host: ['/var/tmp', '/mnt/data', '']
}
```

The array structure allows you to pass in cloudfriend operators without encountering parsing issues.

cc @rclark @rodowi


emilymdubois:
Tested an ephemeral volume mount with `cloudfriend.sub` and `cloudfriend.ref` operators and all went 👌

PR #118, opened 2017-02-14
Pin to LTS ubuntu version

springmeyer:
I noticed we are using `ubuntu:latest` in this repo while we consistently pin to a LTS for other ecs projects. This changes to pinning. /cc @rclark any reason not to pin?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu
+FROM ubuntu:14.04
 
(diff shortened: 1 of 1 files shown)


rclark:
Thanks @springmeyer -- dev mode coming back to haunt us. Why not 16.04?

springmeyer:
@rclark - no reason, changed to `16.04`

PR #123, opened 2017-02-16
Make ecs-cluster errors noop and not notify

### LINKED ISSUE #121: Ignore CannotStartContainerError errors
Let's ignore these types of errors, and simply return the message to the queue for another retry.

cc @jakepruitt

jakepruitt:
These errors are noisy and cause @mapsam and @GretaCB to lose sleep. We're better off just handling them within watchbot.

cc/ @rclark @GretaCB 

Closes #121

### DIFF
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -76,8 +76,7 @@ module.exports = function(cluster, taskDefinition, containerName, concurrency, q
     var params = envToRunTaskParams(env, taskDefinition, containerName, startedBy);
 
     ecs.runTask(params, function(err, data) {
-      if (err && (err.code === 'CannotStartContainerError' || err.code === 'CannotPullContainerError' || err.code === 'DockerTimeoutError')) {
-        err = new Error(err.message);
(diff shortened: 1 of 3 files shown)


lib/tasks.js
```diff
@@ -76,6 +76,11 @@ module.exports = function(cluster, taskDefinition, containerName, concurrency, q
     var params = envToRunTaskParams(env, taskDefinition, containerName, startedBy);
 
     ecs.runTask(params, function(err, data) {
+      if (err && (err.code === 'CannotStartContainerError' || err.code === 'CannotPullContainerError' || err.code === 'DockerTimeoutError') {
```
mapsam: ❤️

rclark:
I thought there was some concern over silencing `CannotPullContainerError`s? Will we still have insight to pressure for a better ECR API?

jakepruitt:
@rclark from what I could tell yesterday, I'm still not sure if ECR API is really the cause of those errors. The huge load we're seeing on the EC2's network and disk I/O, I think this problem will require more investigation on our end and ECR's end.

Either way, I definitely think it will be a while before we know a solution, and unpacker shouldn't be getting paged in the meantime.

Issue #128, opened 2017-02-23
Adjust log groups, stream names


rclark:
We want to get gitsha's out of our log group names, so that log groups persist between deploys. Will still want the gitsha involved in the mix, so later on we can deduce what version any set of logs came from.

1. [Remove the serviceVersion](https://github.com/mapbox/ecs-watchbot/blob/3b024444400667f6da4c5219de4dfa302bd96681/lib/template.js#L733) from the log group's name.

2. [Change the stream prefix](https://github.com/mapbox/ecs-watchbot/blob/3b024444400667f6da4c5219de4dfa302bd96681/lib/template.js#L592) so that it includes the gitsha.

cc @mapsam @yhahn

GretaCB:
Happy to lend a hand here, esp as I learn more about metric-ifying

mapsam:
@rclark do we need to also remove the `options.serviceVersion` from the [LogGroupName of the Logs::LogGroup](https://github.com/mapbox/ecs-watchbot/blob/3b024444400667f6da4c5219de4dfa302bd96681/lib/template.js#L188)?

rclark:
Yes! In fact, [here were I originally linked](https://github.com/mapbox/ecs-watchbot/blob/3b024444400667f6da4c5219de4dfa302bd96681/lib/template.js#L733) we should probably be using a `cf.ref` or `cf.getAtt` to grab the log group's name instead of repeating the same thing twice.

mapsam:
> using a cf.ref 

Good idea, on it.

Seeing `awslogs-stream-prefix` in two places as well, for the [worker](https://github.com/mapbox/ecs-watchbot/blob/3b024444400667f6da4c5219de4dfa302bd96681/lib/template.js#L592) and [watcher](https://github.com/mapbox/ecs-watchbot/blob/3b024444400667f6da4c5219de4dfa302bd96681/lib/template.js#L641). Same thing?

rclark:
Yes please!

mapsam:
@rclark can this be closed now that #129 has been merged?

PR #129, opened 2017-02-23
Adjust log groups

mapsam:
Adjusts the log group names to not include the version, which led to the log groups being deleted and recreated on deploys, thus removing manually created logs. To preserve the ability to see which gitsha logs are from, adding gitsha to the `awslogs-stream-prefix` instead.

Refs: #128 

cc @rclark @GretaCB @yhahn

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -589,7 +589,7 @@ module.exports = function(options) {
             Options: {
               'awslogs-group': cf.ref(prefixed('LogGroup')),
               'awslogs-region': cf.region,
-              'awslogs-stream-prefix': cf.join('-', [cf.stackName, options.serviceVersion])
+              'awslogs-stream-prefix': options.serviceVersion
             }
           }
         }
@@ -638,7 +638,7 @@ module.exports = function(options) {
             Options: {
               'awslogs-group': cf.ref(prefixed('LogGroup')),
               'awslogs-region': cf.region,
-              'awslogs-stream-prefix': cf.join('-', [cf.stackName, options.serviceVersion])
+              'awslogs-stream-prefix': options.serviceVersion
             }
           }
         }
@@ -730,7 +730,7 @@ function logAggregator(prefixed, resources, conditions, options) {
     DependsOn: prefixed('LogGroup'),
     Properties: {
       DestinationArn: options.logAggregationFunction,
-      LogGroupName: cf.join('-', [cf.stackName, cf.region]),
+      LogGroupName: cf.ref(prefixed('LogGroup')),
       FilterPattern: ''
     }
   };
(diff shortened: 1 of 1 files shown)


lib/template.js
```diff
@@ -589,7 +589,7 @@ module.exports = function(options) {
             Options: {
               'awslogs-group': cf.ref(prefixed('LogGroup')),
               'awslogs-region': cf.region,
-              'awslogs-stream-prefix': cf.stackName
+              'awslogs-stream-prefix': cf.join('-', [cf.stackName, options.serviceVersion])
```
mapsam: @rclark you mentioned using a `cf.ref` here - but I'm not sure which value to reference, other than just making a reusable variable `var streamPrefix = cf.join('-', [ ... ])`

lib/template.js
```diff
@@ -730,7 +730,7 @@ function logAggregator(prefixed, resources, conditions, options) {
     DependsOn: prefixed('LogGroup'),
     Properties: {
       DestinationArn: options.logAggregationFunction,
-      LogGroupName: cf.join('-', [cf.stackName, cf.region, options.serviceVersion]),
+      LogGroupName: cf.join('-', [cf.stackName, cf.region]),
```
rclark: This is where I think you could use a `cf.ref(prefixed('LogGroup'))` instead of rewriting `cf.join('-', [cf.stackName, cf.region])`

lib/template.js
```diff
@@ -638,7 +638,7 @@ module.exports = function(options) {
             Options: {
               'awslogs-group': cf.ref(prefixed('LogGroup')),
               'awslogs-region': cf.region,
-              'awslogs-stream-prefix': cf.stackName
+              'awslogs-stream-prefix': cf.join('-', [cf.stackName, options.serviceVersion])
```
rclark: I don't think there's any purpose in including the stackName again here. To find a logstream, you have to go through a loggroup. If the log group has the stack's name in it, then you should know, right?

lib/template.js
mapsam: Ah, I see. Didn't realize `LogGroup` was the equivalent of the stackname+region.

lib/template.js
mapsam: That sounds correct to me - I'll go with just the version and can paste some images of what that looks like in action.

rclark:
I'm going to merge this now.

This will be a breaking change and a tricky rollout on existing stacks due to log group naming conflicts. However every new stack that we roll out without this change only expands the number of tricky rollouts we'll need to do later.

mapsam:
Cool - thanks @rclark!

PR #137, opened 2017-03-14
Collect metrics on pending task times

rclark:
This adds another watchbot metric to record the time between a task's `createdAt` and `startedAt` times.

cc @jakepruitt


jakepruitt:
This looks good! Gonna release and deploy to unpacker (and ecs-conex too).

PR #146, opened 2017-04-16
Broadcast FailedWorkerPlacement metric under the name of the cluster

emilymcafee:
Sends the FailedWorkerPlacement metric to CloudWatch under the name of the selected cluster in addition to the name of the stack. This allows you to tune cluster scaling policies to react to a failure of the watcher to place worker containers on the cluster.

cc @rclark - have we thought about this? Is this a good idea?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+### 2.4.0
+
+- Broadcast FailedWorkerPlacement metric under the name of the cluster, in addition to stack name.
+
 ### 2.3.1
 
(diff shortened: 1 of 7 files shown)


emilymcafee:
\o/ Turns out [we already](https://github.com/mapbox/ecs-cluster/blob/master/cloudformation/ecs-cluster.template.js#L724-L739) [do this](https://github.com/mapbox/ecs-cluster/blob/c8f4d83b1f3b787e8a9a2a47d078e6c7029493f0/lib/metrics.js#L222-L241)!

PR #163, opened 2017-09-14
Add clear definition for message timeout parameter

amishas157:
This PR intends to add clearer definition for `messageTimeout` parameter in building template doc as the messageTimeout is not actually the time after which workers times out and exits, but rather it is the time when this message is set to be invisible for any other consumers. But after this timeout, message becomes available again and is processed by worker.

https://github.com/mapbox/ecs-watchbot/blob/24eee9f6c6e44b97132e1c28b2f3aa8329b2dc1e/lib/template.js#L231

This came to me because I was always in the belief that worker exits after hitting messageTimeout limit, which is not true.

Also, feel free to edit it as it might still sound fuzzy.

cc @arunasank @emilymcafee


jakepruitt:
We've removed the `messageTimeout` option in watchbot 4. Closing this for now.

Issue #169, opened 2017-10-19
Make the failed worker placement alarm customizable


nickcordella:
Would it be possible to make the `failedWorkerPlacementAlarm` more customizable by accepting optional parameters for [some of the properties](https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L301-L303), as is done with the [queue size alarm](https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L406-L407), for instance?

This is handy if one is causing regular, periodic, large cluster scale-ups that cause transient spikes in the failed placements. I am happy to take a crack at a PR. I'd probably make the `EvaluationPeriods` an option, and maybe the `Threshold` too.

cc @rclark

rclark:
Sounds good @nickcordella

oini:
Implemented in PR #170

PR #170, opened 2017-10-23
Make failed worker placement alarm periods customizable

oini:
Ref #169 

This PR makes the `EvaluationPeriods` for `FailedWorkerPlacementAlarm` customizable

cc @rclark @nickcordella

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -120,6 +120,12 @@ var table = require('@mapbox/watchbot-progress').table;
  * triggering an alarm. You specify the number of 5-minute periods before an
  * alarm is triggered. The default is 24 periods, or 2 hours. This parameter
  * can be provided as either a number or a reference, i.e. `{"Ref": "..."}`.
+ * @param {number|ref} [options.failedWorkerPlacementAlarmPeriods=1] - Use this
+ * parameter to control the duration that the SQS queue must be over the failed
+ * worker placement threshold before triggering an alarm. You specify the number
+ * of 1-minute periods before an alarm is triggered. The default is 1 period, or
+ * 1 minute. This parameter can be provided as either a number or a reference,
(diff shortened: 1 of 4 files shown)


lib/template.js
```diff
@@ -120,6 +120,12 @@ var table = require('@mapbox/watchbot-progress').table;
  * triggering an alarm. You specify the number of 5-minute periods before an
  * alarm is triggered. The default is 24 periods, or 2 hours. This parameter
  * can be provided as either a number or a reference, i.e. `{"Ref": "..."}`.
+ * @param {number|ref} [options.failedWorkerPlacementAlarmPeriods=1] - Use this
```
nickcordella: This is a minor point, but how about making the name a bit shorter - maybe get rid of "Worker". Just `failedPlacementAlarmPeriods`?

test/template.test.js
```diff
@@ -41,6 +41,7 @@ test('[template] bare-bones, all defaults, no references', function(assert) {
   assert.ok(watch.Resources.WatchbotTaskEventRule, 'task event rule');
   assert.ok(watch.Resources.WatchbotTaskEventQueuePolicy, 'task event queue policy');
   assert.ok(watch.Resources.WatchbotFailedWorkerPlacementMetric, 'failed worker metric');
+  assert.ok(watch.Resources.WatchbotFailedWorkerPlacementAlarm, 'failed worker alarm');
```
nickcordella: In addition to making sure the alarm exists, can we also make sure that the configurable `EvaluationPeriods` is also set properly? See line 51 with the `watch.Resources.WatchbotWorkerErrorsAlarm.Properties.Threshold` assertion as an example

oini:
Made all the requested changes @nickcordella 

@rclark Would love to get a 👍  from you before merging

PR #173, opened 2017-11-08
Log messages with missing containers

GretaCB:
Investigating why a task would stop and lead to a cloudwatch event where `containers` is an empty array.

cc @rclark @mapsam


jakepruitt:
Watchbot 4 has a very different mechanism for doing work, so I think we can remove this logging message for now.

If this error persists, we can add debugging logs as needed.

Issue #175, opened 2017-12-12
Problems when watcher receives SQS messages 2nd+ time


rclark:
There are situations where an SQS message might be seen by the watcher more than once. Two example cases:

- Watcher gets message A, launches container A1. The SQS message timeout is configured to 10 minutes. After this timeout passes, container A1 is still running. SQS drops message A back into the queue, and the watcher picks it up again.

- Watcher gets message B, makes API request to launch container B1. ECS control plane fails to communicate with the ecs-agent on some EC2, leaving the task in PENDING state. After the SQS message timeout passes, the watcher still thinks container B1 is running. SQS drops message B back into the queue, and the watcher picks it up again.

The watcher keeps an index of message IDs that it considers "in-flight". These are SQS messages that the watcher thinks it has spawned jobs for. It uses this index to filter through cloudwatch events to learn how the containers it cares about exited (success vs failure?), and react by either returning the SQS message to the queue for a retry, or removing it.

In these two cases, the watcher receives and SQS message whose ID is already in the in-flight index. In this case, **the watcher silently ignores the message**.

https://github.com/mapbox/ecs-watchbot/blob/175038669d5cdc4115d9f4ad7aa4464ea630f1cb/lib/messages.js#L86-L90

In the code here, `envs` represents the SQS messages that have just be retrieved. It just tosses out the messages that it thinks are in-flight, and passes the others on to be launched as ECS tasks.

## Impact

- In the case of message A and container A1, this might be a feature -- no need to launch another container, since container A1 is still working on the job.
- In the case of message B, there is no container working on the job and this is a bug -- we need to get a container launched to process the message.
- Either way, the watcher has just received a message and it does nothing with it. This counts as a receive (10 receives put a message into the dead letter queue). There is no log to indicate that the message was received. It does not return the message to the queue, which means the message will just sit around until it hits the SQS message timeout again.

If you're reading watcher logs, depending on the timeouts and container run durations, this can make it look like a message was only received once, and then ended up in the dead letter queue.

## Resolution

I'm hesitant to change the behavior, and let the watcher launch two of the same task. However, I think its important that the watcher log every message that it received from SQS **and also** logs for each message that it launches a container for. Its also important that SQS messages that it no-ops on get properly returned to SQS, with appropriate backoff based on the number of receives.

[...]

PR #176, opened 2017-12-13
Log SQS duplicate receives, and attempt to stop long-pending tasks

rclark:
This PR adds logging of SQS messages that are received multiple times while a previous task is already in flight.

This is only half of what I said I'd implement in #175, but I'm second-guessing myself on the the other half, and will follow up in that ticket.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -133,7 +133,23 @@ module.exports = function(config) {
               })
             );
 
-            queue.defer(tasks.stopIfPending, skipped);
+            queue.defer(function(next) {
+              tasks.stopIfPending(skipped, function(err, stopped) {
+                if (err) return next(err);
+
+                if (stopped) log.info(
+                  '[%s] [stopped-pending] %s',
+                  skipped.MessageId,
+                  JSON.stringify({
+                    subject: skipped.Subject,
+                    message: skipped.Message.substr(0, 2048),
+                    receives: skipped.ApproximateReceiveCount
+                  })
+                );
+
+                next();
+              });
+            });
           });
         }
 
--- a/lib/main.js
+++ b/lib/main.js
@@ -118,23 +118,33 @@ module.exports = function(config) {
         log.debug('[debug] %s messages in flight: %s', Object.keys(messages.inFlight).length, Object.keys(messages.inFlight).join(', '));
         log.debug('[debug] messages received: %s', envs ? envs.length : 0);
 
-        // Log duplicate receives
-        if (skips) skips.forEach(function(skipped) {
-          log.info(
-            '[%s] [duplicate-receive] %s',
-            skipped.MessageId,
-            JSON.stringify({
-              subject: skipped.Subject,
-              message: skipped.Message.substr(0, 2048),
-              receives: skipped.ApproximateReceiveCount
-            })
-          );
+        var queue = d3.queue(10);
+
+        // Log duplicate receives, and stop in-flight tasks if they are pending
+        if (skips) {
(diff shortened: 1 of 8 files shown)


rclark:
Now this adds new functionality:

- If the watcher receives a duplicate task, log it.
- Then, make an ECS DescribeTasks API call to determine whether the previous task is in `PENDING` state.
- If it is, make an ECS StopTask API call to try and stop the pending task, log that this happened.

  At this point, the watcher will have be a position where it is waiting to receive a CloudWatch event indicating that, in fact, the pending task has been stopped without completing successfully. At that point, it will return the SQS message to the queue, and pick it up and attempt it again later (or on another watcher).

- If the task is NOT in `PENDING` state when the DescribeTasks call is made, the watcher does nothing. It is making the assumption that the task is running smoothly, just taking a little longer than expected. 

Next up: trying to figure out how to prove this work in a staging environment.

lib/main.js
```diff
@@ -111,16 +111,56 @@ module.exports = function(config) {
...
         if (err) log.error(err);
 
         // If there are no messages, wait a second before repeating
         log.debug('[debug] %s messages in flight: %s', Object.keys(messages.inFlight).length, Object.keys(messages.inFlight).join(', '));
         log.debug('[debug] messages received: %s', envs ? envs.length : 0);
-        if (!envs || !envs.length)
-          return setTimeout(main, 1000, config, emitter);
 
-        messagesPolled(envs);
+        var queue = d3.queue(10);
```
jakepruitt: why not promises here?

lib/messages.js
```diff
@@ -83,15 +83,17 @@ module.exports = function(queue, topic, stackName, sendNotifications, logGroup)
       if (!data.Messages || !data.Messages.length)
         return callback(null, []);
 
+      var skips = [];
       var envs = data.Messages.filter(function(message) {
-        var notInFlight = !messagesInFlight[message.MessageId];
+        var isInFlight = messagesInFlight[message.MessageId];
+        if (isInFlight) skips.push(messageToEnv(message));
```
jakepruitt: Do skips ever get removed from the `skips` array?

rclark: `.poll()` only gets called once per watcher cycle. Each `.poll()` call = `1` receiveMessage request to SQS. Unless I'm missing something, `skips` isn't a variable that persists between subsequent cycles, so it doesn't need to get emptied.

jakepruitt: 👍 gotcha - makes sense.

lib/main.js
rclark: Wrote watchbot before we started writing all our I/O with promises. When in Rome...

jakepruitt: 👍 romans sound like a old school bunch.

rclark:
Had some successful experiments with this today. The only catch was that a `stopTask` API call on a task that's `PENDING` _might not always_ trigger a CloudWatch event that the watcher can wait for as indication that the task actually stopped. There can be a significant delay in ECS finding out that the task was successfully stopped, and the reasons for this have a lot to do with why the task was probably pending in the first place.

Instead, if it made a successful `stopTask` API call, the watcher will just act like it also received the event indicating that the task has actually stopped. This means the SQS message is returned to the queue as soon as the old `PENDING` task was identified and requested to be stopped.

rclark:
@nickcordella @amishas157 heads up that I published v3.5.0 with this change. It'd be great if you could deploy it to your processing stacks that have been wrestling with pending-task issues over the last week or so.

PR #177, opened 2018-01-03
Adding us-west-1 support to watchbot

JackDanger:
Prior to this change creating a new stack via cfn-config in us-west-1 failed with the following error:
```
Failed to create stack: ValidationError: Template error: Unable to get mapping for EcrRegion::us-west-1::Region
```

I've enabled support for just this one region if we'd like to make only minor changes. There's an alternative PR [here](https://github.com/mapbox/ecs-watchbot/pull/178) that adds every region.

### DIFF
--- a/mappings/ecr-region.json
+++ b/mappings/ecr-region.json
@@ -2,6 +2,9 @@
     "us-east-1": {
         "Region": "us-east-1"
     },
+    "us-west-1": {
+        "Region": "us-west-1"
+    },
     "us-west-2": {
         "Region": "us-west-2"
     },


JackDanger:
The test suite is failing because two tests in `template.test.js` require live access to AWS and are failing for lack of credentials. I'd be interested in adding [Moto Server](https://github.com/spulec/moto) support to the test suite to mock out AWS entirely unless there's a better solution that the team has in mind.

rclark:
https://github.com/mapbox/ecs-watchbot/blob/d1dda99b07736431365bdc58116374206cbd0687/test/template.test.js#L435

Both of the failing tests are situations where watchbot is asked to construct a template body, and then the test script hands it off to [ValidateTempate API](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_ValidateTemplate.html) to prove that watchbot generated a valid template. If we were to mock the AWS API here, unfortunately we wouldn't really be accomplishing what the test sets out to accomplish.

That said, the validation performed by that API call is not really complete -- a template that can't be deployed can still make it through this API call ok, though the API can catch some more glaring errors. 

All this is to say, please don't sweat these test failures. I can run them locally to confirm that nothing is busted, and then we can merge the PR -- I think we just want to figure out whether to merge this one, or #178. Do you have thoughts on that?

JackDanger:
> All this is to say, please don't sweat these test failures.

Thanks, @rclark! I was worried that I'd just opened a failing PR that added a net negative value to your team.

> I think we just want to figure out whether to merge this one, or #178. Do you have thoughts on that?

Thank you for the history on this, it makes plenty of sense now. I've adjusted #178 to be purely additive (i.e. new regions are supported and pull images from ECR in existing regions) and I've removed the compliance-restricted regions (us-gov-west-1, cn-north-1, cn-northwest-1). My (admittedly still totally ignorant) recommendation would be to now merge #178. I believe it'll unblock Watchbot users attempting to deploy to several regions.

Thanks for the help on this!

JackDanger:
Closing in favor of #178

PR #178, opened 2018-01-05
Add every* AWS region to Watchbot ECR support

JackDanger:
[* not adding GovCloud or China]

Because there's no Cloudformation support to provide a default value to `Fn::FindInMap` we must explicitly create a mapping for each AWS region in our ECR region map.
I've added the missing regions and for compliance-restricted locations (GovCloud, China) the mapping points to the given region (GovCloud -> GovCloud, China -> China).
For all other regions the mapping points to the geographically nearest region that already has Watchbot ECR support.

If this seems reasonable then this PR supercedes [this PR](https://github.com/mapbox/ecs-watchbot/pull/177) that only added one US region.

### DIFF
--- a/mappings/ecr-region.json
+++ b/mappings/ecr-region.json
@@ -11,7 +11,19 @@
     "eu-west-1": {
         "Region": "eu-west-1"
     },
+    "eu-west-2": {
+        "Region": "eu-west-2"
+    },
+    "eu-west-3": {
+        "Region": "eu-west-3"
+    },
     "ap-northeast-1": {
+        "Region": "us-west-2"
+    },
+    "ap-northeast-2": {
+        "Region": "us-west-2"
+    },
+    "ap-south-1": {
         "Region": "us-west-2"
     },
     "ap-southeast-1": {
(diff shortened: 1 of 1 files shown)


rclark:
The history here is that this library was developed alongside https://github.com/mapbox/ecs-conex. That library was the first example of ecs-watchbot in action, and its purpose is to build docker images and upload them to ECR when commits are made to a github repo.

At the time, ECR was only available in 3 regions. ecs-conex pushes images to ECR in those 3 regions, and so if an ecs-watchbot stack is about to launch in another region (us-west-1 for example), then it'd need to grab the image from one of the 3 regions where ECR existed.

Nowadays, ECR is in most or all regions. We could do away with the mapping, and get ecs-conex writing to more regions, or we could keep the mapping -- but we have to keep the ECR regions we read from consistent with where ecs-conex will write images. Right now that doesn't include the restricted or china regions.

There are other headaches that'd be involved in getting ecs-conex to write to restricted and china regions, so maybe we ought to leave those out of this PR for now?

jakepruitt:
Mapping file is no longer needed.

PR #186, opened 2018-03-13
Move binary-split from dev to dependency

brendanmcfarland:
Ran into `Error: Cannot find module 'binary-split'` when running `watchbot.log` on this branch, thinking this module needs to be a dependency.

cc: @rclark

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -563,7 +563,6 @@
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/binary-split/-/binary-split-1.0.3.tgz",
       "integrity": "sha1-yqiKyNhZ0zFw9fHOa0xZCHn3UCY=",
-      "dev": true,
       "requires": {
         "through2": "2.0.3"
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -7020,8 +7018,7 @@
     "xtend": {
       "version": "4.0.1",
       "resolved": "https://registry.npmjs.org/xtend/-/xtend-4.0.1.tgz",
-      "integrity": "sha1-pcbVMr5lbiPbgg77lDofBJmNY68=",
-      "dev": true
+      "integrity": "sha1-pcbVMr5lbiPbgg77lDofBJmNY68="
```
rclark: I'm noticing a lot of issues in package-lock.json where one person does an install and gets a `sha1-`, another gets a `sha512-`. I found this issue: https://github.com/npm/npm/issues/16938#issuecomment-350397663 and I wonder if we need to get everyone to upgrade npm.

PR #187, opened 2018-03-14
Scale down threshold

brendanmcfarland:
Was getting an error from the scale down alarm

```"error": "No step adjustment found for metric value [0.0] and breach threshold 1.0",```

I think this is due to the lower bound being set at 0.0. Reading step adjustment docs, [a threshold of 1 and lowerbound of 0.0 ](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_StepAdjustment.html) is effectively only a policy for thresholds at `1.0` or above. This pr adjusts the LowerBound to include  `0.0`.

We could do this a billion ways (I think a MetricIntervalUpperBound of 0.0 would also work); I'm pretty agnostic on the cleanest way to define the step adjustment.

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -403,7 +403,7 @@ module.exports = (options = {}) => {
         StepAdjustments: [
           {
             ScalingAdjustment: -100,
-            MetricIntervalLowerBound: 0.0
+            MetricIntervalLowerBound: -1.1
           }
         ]
       }
--- a/lib/template.js
+++ b/lib/template.js
@@ -403,7 +403,7 @@ module.exports = (options = {}) => {
         StepAdjustments: [
           {
             ScalingAdjustment: -100,
-            MetricIntervalLowerBound: -1.1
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/template.js [hunk 1]
```diff
@@ -403,7 +403,7 @@ module.exports = (options = {}) => {
         StepAdjustments: [
           {
             ScalingAdjustment: -100,
-            MetricIntervalLowerBound: 0.0
+            MetricIntervalLowerBound: -1.1
```
rclark: Yeah I think this should be a `MetricIntervalUpperBound` of `0`. 

The metric is `ApproximateNumberOfMessagesNotVisible`. When the metric value is < 1, the alarm triggers, and invokes the scaling policy. The breach value should be -1 at that point. Any breach value less than zero is supposed to trigger this step adjustment.

brendanmcfarland: 👍

PR #188, opened 2018-03-17
[WIP] Disk monitoring and quotas

jakepruitt:
## Background

The "♻️ that container" pull request #184 adds sizeable improvements to running high concurrency workloads without failed task placements by running multiple jobs in the same container. One major drawback of running multiple jobs in the same container is the potential buildup of files each worker would write and potentially not clean up - either out of developer missteps or hard-failing jobs.

While a disk buildup would have limited consequences in a single-tenant situation where the EC2 was only running one watchbot stack, we don't run in a single-tenant world - our EC2's run many diverse watchbot stacks with varying sensitivities to disk space. With the potential risk of one watchbot stack taking down an entire cluster of EC2 disk space, we can't merge #184 without first having some sort of safeguard in place.

Building a safeguard system for runaway disk writes could come in many shapes and colors, from passively monitoring and aborting when the disk space is too high, to restricting write access to a limited set of directories, to wiping certain directories after every job. This PR begins by trying to do all 3 of those. Before we merge this PR, we will need to make a few decisions (see **Open questions** below) and then implement those decisions in a way that's more in line with watchbot's new code patterns that what I've written.

It's worth mentioning here that volume driver options are on [ECS's mind](https://github.com/aws/amazon-ecs-agent/issues/236) and may allow us to make most of this code obsolete. Specifically if we design around inherently limited volumes and an inability to write to other filepaths, we will be able to work pretty well with volume driver options.

## Design thoughts

As a draft for what needs to be added to the codebase I think in general:

1. At watcher upstart, we need to parse options from the template, like volume paths, an `autoClean` or `persist` boolean option, a `maxDiskSpace` parameter and potentially other parameters.
2. If we want to actively poll the filesystem for usage information, we will need to kick off this polling at the very beginning.
3. If we need to create a user role that has limited access to the filesystem, we will need to create it at the very beginning and set the volume directories as explicitly writable.
4. If we want to initialize the volumes with built-in maximums (refs https://stackoverflow.com/questions/8148715/how-to-set-limit-on-directory-size-in-linux) we need to do that at watcher startup.
5. When we start worker sub-processes, we will need to start them with the user ID or group ID that unable to write anywhere but the volumes.
6. After the worker is complete, regardless of status code, we should clean the `tmp` directory and the volume drives if `autoClean` is specified to true.

## Testing locally and iterating

I was able to test the branch by deploying a watchbot stack with `workers: 0` and running `QueueUrl=<Queue url from deployed watchbot stack> ./bin/watchbot listen 'echo "hello"'`. I'm still in the middle of debugging problems - right now I still don't know if the filesystem polling is working, and it looks like the `gid: 100` wasn't working for restricting file writes successfully.

We still need to answer a bunch of higher-level questions before digging more into the code - I think most of the code so far is ill-planned, but from this initial cut most of the things we're thinking about seem technically possible to fit into watchbot's internal architecture.

## Open questions 🤔 

- **Preventing writes vs. allowing writes and monitoring** - There may be some tough developer friction with preventing any writes to non-volumes, although it would give us a clearer path for knowing that after each run we would be cleaning the only paths that containers could write to. We could decide to let users continue to write anywhere, and promise to clean volumes and `/tmp` after each job. A regular polling monitor could detect when the system is overfilled and quit, giving an appropriate error message when that happens.
- **Quota everything vs. just quota volumes** - We should decide whether we want the quota-ing system on only the volumes or on the entire filesystem. (The watchbot PR initially checks system-wide.) If we prevent writes to anywhere except the volumes, quotaing the entire filesystem would be redundant.
- **Debugging prevented writes** - How will local testing work for debugging filesystem issues? For instance if there's a program in the container that needs access to a file that's not granted to users, how will developers be able to try out different `chmod`'s without deploying?
- **Cleaning /tmp** - Should we automatically wipe the `/tmp` directory after every worker? I feel like we should do this just from the nature of the `/tmp` directory.
- **Allowing `persist` mode** - Some developers may appreciate the new value of sometimes having access to files created by the last worker, and we may want to provide a template parameter for keeping the files around after each job.
- **Deciding on default quota size** - Out of the gates I think we should make it large enough to handle most use cases - somewhere around 100GB or potentially tighter around 50 GB. Enough to prevent people from shooting themselves in the foot.

## Next actions

- [x] Come up with conclusions about the above questions
- [ ] If we want to prevent writes, we need to figure out what user or group id would have read-only permissions on all of the filesystem - I was hoping we could do this with a default linux group ID, but we may need to create a user at worker upstart. It seems like `gid: 100` didn't work in my initial attempts at running an `echo` command - it gave errors about permissions even with that.
- [ ] Test out `mount`-related commands on the volumes (refs https://stackoverflow.com/questions/8148715/how-to-set-limit-on-directory-size-in-linux) to see if they will work for quota-ing the volume sizes.
- [ ] Build the parameters for switching between "perisist" mode and "autoClean" modes so users can specify whether they want watchbot to keep files in volumes rather than cleaning them.

cc/ @brendanmcfarland @tapasweni-pathak @rclark @yhahn @arunasank

### DIFF
--- a/lib/file-system.js
+++ b/lib/file-system.js
@@ -1,5 +1,9 @@
 'use strict';
 
+const rmdir = require('fs').rmdir;
+const chmod = require('fs').chmod;
+const spawn = require('child_process').spawn;
+
 class FileSystem {
   /**
    * @name Watcher
@@ -10,6 +14,7 @@ class FileSystem {
   constructor(options = { volumes: [], autoClean: true }) {
     this.autoClean = options.autoClean;
     this.volumes = options.volumes;
+    this.maxDiskSpace = options.maxDiskSpace;
   }
 
   clean() {
@@ -26,6 +31,24 @@ class FileSystem {
     } else {
       return Promise.resolve();
     }
+  }
+
+  checkDisk() {
+    return new Promise((resolve, reject) =>
+      let output = '';
+
+      spawn('du', ['-s', '/'])
+        .on('error', (err) => reject(err))
+        .on('exit', (code, signal) => {
+          console.log(output);
+          const match = output.match(/(\d+)\s+\//);
+          const size = Number(match[1]);
+
+          if (size > this.maxDiskSpace) resolve('full');
+          resolve();
+        })
+        .stdout.on('data', (data) => output.concat(data));
+    );
   }
 
   init() {
--- a/lib/file-system.js
+++ b/lib/file-system.js
@@ -1,0 +1,47 @@
+'use strict';
+
+class FileSystem {
+  /**
+   * @name Watcher
+   * @param {Object} options - configuration
+   * @param {Array} options.volumes - array of container paths for docker volumes
+   * @param {Boolean} options.autoClean - whether or not to delete files after every worker run. default true.
+   */
+  constructor(options = { volumes: [], autoClean: true }) {
+    this.autoClean = options.autoClean;
+    this.volumes = options.volumes;
+  }
+
+  clean() {
+    if (this.autoClean) {
+      const rms = this.volumes.concat(['/tmp']).map((volume) =>
+        new Promise((resolve, reject) => {
+          rmdir(volume, (err) => {
+            if (err) return reject(err);
+            resolve();
+          });
+        });
+      );
+      return Promise.all(rms);
+    } else {
+      return Promise.resolve();
+    }
+  }
+
+  init() {
+    const chmods = this.volumes.map((volume) =>
+      new Promise((resolve, reject) => {
+        chmod(volume, 0o777, (err) => {
+          if (err) return reject(err);
+          resolve();
+        });
+      });
+    );
+
+    await Promise.all(chmods);
+  }
+
+  static create(options) {
+    return new FileSystem(options);
+  }
+}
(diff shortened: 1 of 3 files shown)


jakepruitt:
Further experiments to try:

- Exiting 1 (killing the container) when job fails.
- Exiting 0 (killing the container) after every job.
- Create a hard ceiling on disk volume space.
- Create a watchbot metric for disk usage per task, both in absolute numbers and percentage.
- Create user limitation so subprocess can only write to volumes

jakepruitt:
## The plan

After some deliberation, we decided on a few things for the disk management system in a recycled container environment:

1. **Preventing writes vs. allowing writes and monitoring:** We should **prevent writes to most of the file system except docker volumes** and **always make the `/tmp` directory a docker volume.**
2. **Quotas and monitoring** We decided to go with the restricted writes and a clean sweep of the volumes after every container run. Quotas and closer monitoring may be useful in the future, but first we should focus on restriction and cleaning.

In terms of this PR, we need to implement a few things:

- [ ] Figure out how to run the subprocess with a user that only has access to write to docker volume paths. So far, I think this can be done using a group ID for the subprocess and won't require writing code to create a new user, but we may need to run `useradd` at some point when the watchbot worker starts up to add a worker user that can only write to certain areas.
- [ ] Remove all of the files in the docker volumes after every job completion.

jakepruitt:
Closing this for now as an experiment, and moving in the direction of #197 as a stepping stone to get to this point. Will be opening up a new PR to address #197.

Issue #190, opened 2018-04-23
[experiment] Service-based self-polling watchbot with containers that exit after every job


jakepruitt:
## The background

Helpful reading:

- ♻️ that container pull request: https://github.com/mapbox/ecs-watchbot/pull/184
- Disk monitoring and quotas: https://github.com/mapbox/ecs-watchbot/pull/188
  - "With the potential risk of one watchbot stack taking down an entire cluster of EC2 disk space, we can't merge #184 without first having some sort of safeguard in place."

The current implementation of #184 "♻️ that container" runs the risk of filling disk space on ECS hosts as successive watchbot jobs are run inside the same container. In order to prevent the hosts from filling up with the leftover files of every job, we could potentially try a few different options:

- Stop containers after every job, thereby creating a fresh new environment for every job to run in. This defeats some of the goals of the "♻️ that container" pull request, but it still leaves us in a place that doesn't rely on the RunTask API. It will instead rely on the service scheduler, which may or may not have the same underlying throughput problems as the RunTask API. We cannot know the behavior until we test it out.
- Build code into the watcher that restricts the write access of the worker subprocess to just the docker volumes, then wipe the docker volumes clean after every job.

This ticket outlines the plan for the first branch of the experiment: making every container exit after every job. 

## The plan

We should start with the #184 "♻️ this container" branch's code. @rclark has already done the legwork of writing code for running watchbot as an ECS service, polling the SQS queue from within the container, and running the main script as a subprocess.

The code we'll want to change is specifically around the way the watcher handles the worker's success - we'll want to either `process.exit()` the main loop after the worker completes, or do something that's easier to mock/test. I think doing `process.exit()` is probably fine for now, and if we decide this is the direction we want to pursue, we can refine the code a bit more.

## The next actions

Basically, our next actions are to implement the plan. I think this can be done with some specific code changes:

- [ ] Make a branch from the #184 "♻️ this container" branch.
- [ ] Modify the `watcher.js` code to exit the main node process after workers complete.
- [ ] Implement tests that the process exits: there may be some tough mocking to make this easily testable.
- [ ] Confirm the new watchbot version works by testing out with https://github.com/mapbox/ecs-telephone.
- [ ] Load test the new watchbot version with a really high concurrency load of very fast tasks.

cc/ @mapbox/platform-engine-room

arunasank:
Closed via https://github.com/mapbox/ecs-watchbot/pull/196

PR #196, opened 2018-05-18
Fresh mode: Gracefully exit from the container when a batch of jobs finish processing.

arunasank:
Fixes https://github.com/mapbox/ecs-watchbot/issues/190 when done.

# TODO
* [x] Exit after workers are done running in `watcher.js`
* [x] Write tests to mock the entire workflow and make sure that `process.exit()` is called at the end of each container run.
* [x] Confirm that this code runs on ecs-telephone
* [x] Load test with a watchbot stack with high concurrency and fast tasks to look at failed task placement metrics

### DIFF
--- a/lib/message.js
+++ b/lib/message.js
@@ -74,7 +74,6 @@ class Message {
   async complete() {
     const params = { ReceiptHandle: this.handle };
     try {
-      this.logger.log('Message finished processing');
       return await this.sqs.deleteMessage(params).promise();
     } catch (err) {
       this.logger.queueError(err);
--- a/lib/message.js
+++ b/lib/message.js
@@ -74,7 +74,7 @@ class Message {
   async complete() {
     const params = { ReceiptHandle: this.handle };
     try {
-      this.logger.log('[message]: Message finished processing');
+      this.logger.log('Message finished processing');
       return await this.sqs.deleteMessage(params).promise();
     } catch (err) {
       this.logger.queueError(err);
--- a/lib/message.js
+++ b/lib/message.js
@@ -74,6 +74,7 @@ class Message {
   async complete() {
     const params = { ReceiptHandle: this.handle };
     try {
+      this.logger.log('[message]: Message finished processing');
       return await this.sqs.deleteMessage(params).promise();
     } catch (err) {
       this.logger.queueError(err);
--- a/bin/watchbot.js
+++ b/bin/watchbot.js
@@ -14,7 +14,7 @@ const main = async () => {
 
   const options = {
     queueUrl: process.env.QueueUrl,
-    fresh: process.env.fresh,
+    fresh: process.env.fresh === 'true' ? true : false,
     workerOptions: { command }
   };
 
--- a/bin/watchbot.js
+++ b/bin/watchbot.js
@@ -24,7 +24,7 @@ const main = async () => {
   } catch (err) {
     logger.log(`[error] ${err.stack}`);
   }
-  console.log('exiting now. bye!');
+  logger.log('exiting now. bye!');
 };
 
 module.exports = main;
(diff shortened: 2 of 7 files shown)


arunasank:
dzwth5eu8aab9sn

## Let's go natural

Since we want to exit with a `0` status code after the first batch of workers finish running, it might make sense to help them to reach this exit state naturally, instead of interjecting an explicit `process.exit(0)`. 

The `process.exit(0)` that was added has therefore been replaced with a simple `resolve()` to allow the program to exit on its own terms and pace. 🙂 

## Mocks ✅ 

All existing tests have been modified to account for this new behaviour where a watchbot container will process a single batch of messages, at most.

## Next actions
* [ ] Test the `resolve()` instead of `process.exit(0)` behaviour on ecs-telephone.
* [ ] Run a longer concurrency test with tasks that start and stop very quickly to see the impact on the scheduler, and the rate of failed task placements.

arunasank:
## Interesting notes

In the words of @jakepruitt: 
**If a Promise object is not resolved or rejected, it causes node to barf.**

### ☮️ show-me-the-code wisdom, also from @jakepruitt who tested this:
```
const test = require('tape');

function barf() {
  return new Promise((resolve, reject) => {
    console.log('barf');
  });
}

test('test the barf', (assert) => {
  barf().then(() => console.log('does this ever get called?'))
    .catch(() => console.log('how about this?'))
    .then(() => {
      assert.end();
    });
});
```

```
# test the barf
barf
not ok 1 test exited without ending
  ---
    operator: fail
    at: process.<anonymous> (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/index.js:90:19)
    stack: |-
      Error: test exited without ending
          at Test.assert [as _assert] (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/lib/test.js:225:54)
          at Test.bound [as _assert] (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/lib/test.js:77:32)
          at Test.fail (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/lib/test.js:318:10)
          at Test.bound [as fail] (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/lib/test.js:77:32)
          at Test._exit (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/lib/test.js:191:14)
          at Test.bound [as _exit] (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/lib/test.js:77:32)
          at process.<anonymous> (/Users/jake/Projects/ecs-autoscaling/node_modules/tape/index.js:90:19)
          at emitOne (events.js:116:13)
          at process.emit (events.js:211:7)
  ...

1..1
# tests 1
# pass  0
# fail  1
```

arunasank:
> Test the `resolve()` instead of `process.exit(0)` behaviour on ecs-telephone.

🎉🎉 Stack runs well with ZERO RunTask API calls. 🎉🎉

arunasank:
I ran concurrency tests, and as expected for a large number of jobs that complete in an instant, we still see failed task placements.

arunasank:
In spite of the failed task placements, I still feel good about merging this with master, since we know that the service scheduler cannot protect us from failed task placements, and we'll have to account for this by ensuring enough capacity behind the clusters to account for task placements.

lib/message.js
```diff
@@ -74,6 +74,7 @@ class Message {
   async complete() {
     const params = { ReceiptHandle: this.handle };
     try {
+      this.logger.log('Message finished processing');
```
jakepruitt: I'm guessing we don't want to keep these logs around, right?

arunasank:
> One last fix, before we merge this: This should be enabled on fresh mode alone, and not in all cases.

This was done today, and tested to make sure that the fresh mode is enabled only when `options.fresh` is `true`.

Keep an eye here for notes on the concurrency/task-duration limits for fresh mode.

Issue #197, opened 2018-05-21
Restricting writes to just `/tmp` and cleaning `/tmp` after every worker


jakepruitt:
As a stepping stone for the work that #188 attempted to prove, we should build a system that prevents workers from writing to any directory except the `/tmp` directory and removes all files in the `/tmp` directory after every worker finishes. This system should be built on top of #184.

**Handy links/references**

- Node.js documentation on `child_process.exec`: https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback
- Linux credentials documentation: http://man7.org/linux/man-pages/man7/credentials.7.html
- Linux `useradd` documentation: https://linux.die.net/man/8/useradd
- Linux `groupadd` documentation: http://man7.org/linux/man-pages/man8/groupadd.8.html
- Linux user namespaces documentation: http://man7.org/linux/man-pages/man7/user_namespaces.7.html
- `/tmp` documentation in linux file hierarchy page: http://man7.org/linux/man-pages/man7/file-hierarchy.7.html

cc/ @mapbox/platform-engine-room

jakepruitt:
For restricting writes, I was able to prove that adding a random uid to a spawned child process, whether that uid maps to a real user or not, prevents that child process from writing to anywhere except `/tmp` 🎉 .

Experiment:

<details>

```
const child_process = require('child_process');

const child = async (uid, command) => {
  return new Promise((resolve, reject) => {
    const start = Date.now();

    const options = {
      shell: true,
      env: process.env,
      uid: uid,
      stdio: [process.stdin, 'pipe', 'pipe']
    };

    const child = child_process
      .spawn(command, options)
      .on('error', (err) => reject(err))
      .on('exit', (code, signal) => {
        const duration = Date.now() - start;
        resolve({ uid, code, signal, duration });
      });

    child.stdout.pipe(process.stdout);
    child.stderr.pipe(process.stderr);
  });
};

if (module === require.main) {
  console.log('Before we do anything, let\'s find out some info:');
  console.log(`process.gid: ${process.getgid()}`);
  console.log(`process.groups: ${process.getgroups()}`);
  console.log(`process.uid: ${process.getuid()}`);
  console.log(`process.euid: ${process.geteuid()}`);

  child(200, 'echo "Hello!" > /world-history.txt && cat /world-history.txt && rm /world-history.txt')
    .then((results) => {
      console.log(results);
    }).catch((err) => {
      console.log(err);
    });
}
```

[...]

PR #199, opened 2018-05-22
Restrict writes to tmp and clean after every worker

jakepruitt:
Proves out the ability to restrict writes to the `/tmp` directory, with purging of the `/tmp` contents after every job. Refs #197 

Now that we know it's easy to do this for `/tmp`, we need to find out if it's possible to restrict writes to a list of user-defined volumes. If it seems possible, we should go with that for developer flexibility. If it doesn't seem possible, at least we have this as a fallback now.

## Next actions

- Create a branch off this PR that experiments with restricting writes to user-defined writable volumes. If it seems possible and manageable, let's go with that.

cc/ @mapbox/platform-engine-room @rclark

### DIFF
--- a/test/worker.test.js
+++ b/test/worker.test.js
@@ -134,6 +134,7 @@ test('[worker] waitFor, exit 0', async (assert) => {
 
   sinon.spy(child_process, 'spawn');
   sinon.spy(process.stdout, 'write');
+  sinon.spy(process.stderr, 'write');
 
   try {
     await worker.waitFor();
@@ -154,7 +155,8 @@ test('[worker] waitFor, exit 0', async (assert) => {
     child_process.spawn.calledWith('echo ${Message}', {
       env: Object.assign(message.env, process.env),
       shell: true,
-      stdio: [process.stdin, 'pipe', 'pipe']
+      stdio: [process.stdin, 'pipe', 'pipe'],
+      uid: 200
     }),
     'spawned child process properly'
   );
--- a/test/worker.test.js
+++ b/test/worker.test.js
@@ -3,6 +3,7 @@
 const stream = require('stream');
 const events = require('events');
 const child_process = require('child_process');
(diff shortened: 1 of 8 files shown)


test/worker.test.js
```diff
@@ -171,6 +174,193 @@ test('[worker] waitFor, exit 0', async (assert) => {
...
+  assert.equal(message.complete.callCount, 1, 'called message.complete()');
+
+  Date.prototype.toGMTString.restore();
+  child_process.spawn.restore();
+  process.env = env;
+  logger.teardown();
+  assert.end();
+});
+
+test('[worker] waitFor, /tmp is empty, exit 0', async (assert) => {
```
rclark: Can you clarify what the key assertion in this test is? I'm not sure I understand how it differs from the previous one.

test/worker.test.js
```diff
@@ -171,6 +174,193 @@ test('[worker] waitFor, exit 0', async (assert) => {
...
+  assert.equal(message.complete.callCount, 1, 'called message.complete()');
+
+  Date.prototype.toGMTString.restore();
+  child_process.spawn.restore();
+  process.env = env;
+  logger.teardown();
+  assert.end();
+});
+
+test('[worker] waitFor, attempt to write to root, exit 1', async (assert) => {
```
rclark: per the assertion below, it looks like you end up with exit code 2?

test/worker.test.js
jakepruitt: The goal is to prove that after the previous worker is done, the next worker cannot find any files in that directory. On second thought, I can probably just do this assertion in the first test. I'll do that.

test/worker.test.js
jakepruitt: OOoooo yep good catch.

jakepruitt:
Going to close this in favor of #200, which has a more extensible functionality.

PR #200, opened 2018-05-23
Allow users to write to any volume

jakepruitt:
Adds user-defined volumes to the tmp directory PR #199.

Things I'm not sure about:

- performing `chmod 777` in the watcher before messages begin. Should this be per-worker instead? I figured we might as well save unnecessary work.
- I feel like my tests are missing an important integration piece, but I don't want to actually make `chmod` system calls in the tests. Not sure how I feel there.
- Does this actually work? I'm going to test it out on a quick stack that performs writes to user-defined volumes.
- Code quality: I have one pyramid of doom and a few long-lines in the test code. Should I make those prettier?

Since this isn't a painful implementation (barring any testing surprises), I'm happy to go with this over #199.

cc/ @rclark @mapbox/platform-engine-room

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -2,6 +2,10 @@
 
 const Messages = require('./messages');
 const Worker = require('./worker');
+const fs = require('fs');
+const util = require('util');
+
+const chmod = util.promisify(fs.chmod);
 
 class Watcher {
   /**
@@ -21,8 +25,14 @@ class Watcher {
     this.freshMode = options.fresh;
   }
 
+  async chmodForEveryone(volumes) {
+    return Promise.all(volumes.map((volume) => chmod(volume, 0o777)));
+  }
+
   listen() {
-    return new Promise((resolve) => {
+    return new Promise(async (resolve) => {
+      await this.chmodForEveryone(this.workerOptions.volumes);
+
       const loop = async () => {
         if (this.stop) return resolve();
 
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -2,10 +2,6 @@
 
 const Messages = require('./messages');
 const Worker = require('./worker');
-const fs = require('fs');
-const util = require('util');
-
-const chmod = util.promisify(fs.chmod);
 
 class Watcher {
   /**
@@ -25,14 +21,8 @@ class Watcher {
     this.freshMode = options.fresh;
   }
 
-  async chmodForEveryone(volumes) {
-    return Promise.all(volumes.map((volume) => chmod(volume, 0o777)));
-  }
-
   listen() {
     return new Promise(async (resolve) => {
-      await this.chmodForEveryone(this.workerOptions.volumes);
-
       const loop = async () => {
         if (this.stop) return resolve();
 
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -62,9 +62,7 @@ class Worker {
     const options = {
       shell: true,
       env: Object.assign({}, process.env, this.message.env),
-      stdio: [process.stdin, 'pipe', 'pipe'],
-      // An arbitrarily chosen UID of a non-existent user
-      uid: 200
+      stdio: [process.stdin, 'pipe', 'pipe']
     };
 
     try {
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const child_process = require('child_process');
+const fsExtra = require('fs-extra');
 const Message = require('./message');
 const Logger = require('./logger');
 
@@ -28,6 +29,7 @@ class Worker {
     if (!options.command) throw new Error('Missing options: command');
 
     this.command = options.command;
+    this.volumes = options.volumes;
     this.message = message;
     this.logger = Logger.create('watcher', message);
   }
@@ -52,15 +54,24 @@ class Worker {
     return await this.message.retry();
   }
 
+  async clean(volumes) {
+    return Promise.all(volumes.map((volume) => fsExtra.emptyDir(volume)));
+  }
+
   async waitFor() {
     const options = {
       shell: true,
       env: Object.assign({}, process.env, this.message.env),
-      stdio: [process.stdin, 'pipe', 'pipe']
+      stdio: [process.stdin, 'pipe', 'pipe'],
+      // An arbitrarily chosen UID of a non-existent user
+      uid: 200
     };
 
     try {
       const results = await child(this.command, options, this.logger);
+
+      await this.clean(this.volumes);
+
       if (results.code === 0) return await this.success(results);
       if (results.code === 3) return await this.ignore(results);
       if (results.code === 4) return await this.noop(results);
--- a/test/watcher.test.js
+++ b/test/watcher.test.js
@@ -8,6 +8,7 @@ const Watcher = require('../lib/watcher');
 const Message = require('../lib/message');
 const Messages = require('../lib/messages');
 const Worker = require('../lib/worker');
+const fs = require('fs');
 
 test('[watcher] constructor', (assert) => {
   const messages = stubber(Messages).setup();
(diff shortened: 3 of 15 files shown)


lib/watcher.js
```diff
@@ -20,8 +21,19 @@ class Watcher {
     this.messages = Messages.create({ queueUrl: options.queueUrl });
   }
 
+  async chmod(volumes) {
+    return Promise.all(volumes.map((volume) => {
+      return new Promise((resolve, reject) => {
+        fs.chmod(volume, 0o777, (err) => {
```
rclark: Funtime! try [util.promisify()](https://nodejs.org/docs/latest-v8.x/api/util.html#util_util_promisify_original)!

lib/watcher.js
```diff
@@ -36,6 +48,8 @@ class Watcher {
         setImmediate(loop);
       };
 
+      await this.chmod(this.workerOptions.volumes);
```
rclark: Since the order of these operations is 

1. chmod
2. then start polling

... it would be clearer if you do the `await chmod()` on L24, before the `return new Promise()` step.

lib/watcher.js
```diff
@@ -20,8 +24,14 @@ class Watcher {
     this.messages = Messages.create({ queueUrl: options.queueUrl });
   }
 
+  async chmod(volumes) {
```
vsmart: Could you rename this function to indicate how the permissions are changing here? Without looking up the permissions code `0o777` it's hard to know what this is doing.

vsmart: Edit: Also, is this recursive by default, or do you know to indicate that all files are modified recursively with a `-R`?

lib/worker.js
```diff
@@ -53,8 +54,8 @@ class Worker {
     return await this.message.retry();
   }
 
-  async cleanTmp() {
-    return fsExtra.emptyDir('/tmp');
+  async clean(volumes) {
```
vsmart: Great that you used async/await style! It's very nice to read. 💯

test/watcher.test.js
```diff
@@ -7,6 +7,7 @@ const Watcher = require('../lib/watcher');
 const Message = require('../lib/message');
 const Messages = require('../lib/messages');
 const Worker = require('../lib/worker');
+const fs = require('fs');
```
vsmart: Could you add a test on how this behaves when volumes is empty `[]`? Is there a default volume or does it error out?

jakepruitt:
Note: we _could_ keep persistent volumes around and only clean ephemeral volumes.

jakepruitt:
After digging through the 22 repositories where mounts are defined for watchbot stacks, the only stack that uses persistent volumes is ecs-conex: https://github.com/mapbox/ecs-conex/blob/8014ea9d2e791d54a930c8ad861173d10e911586/cloudformation/ecs-conex.template.js#L22

```
mounts: '/mnt/data:/mnt/data,/var/run/docker.sock:/var/run/docker.sock',
```

While I know the `/var/run/docker.sock` link needs to be there for connecting the two containers, I do not believe the `/mnt/data` needs to be a persistent volume.

For this use case, could we change the `/mnt/data` to a temporary volume?

Could we also change the `/var/run/docker.sock` to a read-only `ContainerDefinition.LinuxParameters.Device`? 

I'll try out ecs-conex with this branch and see what's needed for those modifications.

jakepruitt:
I spent yesterday trying this PR out with https://github.com/mapbox/ecs-conex to see if the restrictions were too limiting for that repository, which is kind of a permissions edge case. After [many different attempts](https://github.com/mapbox/ecs-conex/tree/watchbot4), it seems like the non-sudo user for the child process gets explicitly prevented from using the `/var/run/docker.sock` connection. So I think we'll need to use `fresh: true` (refs #196) for ecs-conex.

A few other notes about this PR:

- In the process of getting ecs-conex set up on watchbot 4, I noticed a few other features in lib/template that needed to be implemented, like returning an object with `refs`. I went ahead and added that here.
- I switched the `Logs` resource to `LogGroup` to maintain backwards consistency
- Since OSX doesn't have a `/tmp` directory, I switched the tests to run in a linux container. This makes the tests a little bit slower to run.
- I'm not 100% certain with how I've created the defaults. I am automatically adding the `/tmp` directory as a volume for all watchbot projects. As far as I know, I don't think this'll cause any problems.

I'm opening this up to review for anyone on @mapbox/platform-engine-room that wants to review it before it merges into #184.

rclark:
> I think we'll need to use fresh: true

I'm worried about this -- both from the perspective on the performance implications that have been raised during benchmarking of #196, and from the fact that root permission feels like a side-effect: you want to run the child process as sudo, and that just so happens to be in the case in "fresh mode" where the primary goal is to turn containers off after every run. 

On the later point I may feel more satisfied if it was `sudo: true` instead of `fresh: true`. At least that makes the contract more clear. "You can run the container as root but you can't run very fast".

jakepruitt:
> On the later point I may feel more satisfied if it was `sudo: true` instead of `fresh: true`. At least that makes the contract more clear. "You can run the child process as root but you can't run very fast".

Just to be extra clear - you're talking about changing the name of `fresh` mode to `sudo` mode, not adding an option in the normal mode where the containers are re-used but you get to run as `sudo`?

If so, I think you're probably right, the `fresh` distinction is probably confusing since normal mode does "freshen" the filesystem. The main difference is the permission you have to write to the entire filesystem or not.

rclark:
I'm teeter-tottering on that distinction.

The feature that you need to make conex work is sudo mode -- where the child process is run as root. The feature that you have is fresh mode, where you allow the child process to write anywhere on the filesystem and you turn of the container after every job. It just so happens that branch is written such that the child process runs as root, though that does seem to be the feature's reason-for-being. 

If you rename fresh -> sudo, the language makes more sense for the conex use case you're solving. However you're adding the baggage of poor scheduler performance and excessive container start/stop overheard where you really don't need that. Conex doesn't want to write anywhere on the filesystem, it just wants to run as root. 

A performant sudo-mode could still use the default logic to clean up a mounted `/tmp` volume and you'd have a performant solution to the conex problem. While I'm not a big fan of the proliferation of "modes" here, I think that sudo-mode on its own is definitely a better solution to the conex problem.

arunasank:
> A performant sudo-mode could still use the default logic to clean up a mounted /tmp volume and you'd have a performant solution to the conex problem. While I'm not a big fan of the proliferation of "modes" here, I think that sudo-mode on its own is definitely a better solution to the conex problem.

Is there a reason why sudo mode cannot be the default in both fresh mode and turbo mode, considering it already is the default in fresh mode?

jakepruitt:
> Is there a reason why sudo mode cannot be the default in both fresh mode and turbo mode, considering it already is the default in fresh mode?

The mechanism that default (turbo) mode uses to restrict writes to only volumes is by changing the user to a non-sudo user and restricting that user's permissions. So as long as we move forward with the restriction approach, we won't implement any sudo behavior in the default mode.

> Conex doesn't want to write anywhere on the filesystem, it just wants to run as root.

> I think that sudo-mode on its own is definitely a better solution to the conex problem

I'm really interested in this kind of use case, and think it may bite us in other projects. It would be nice to be able to say "you're running as root, but root can't write anywhere." Maybe we could acheive this by performing multiple chmod's on all non-volume directories, preventing the root user from ever writing there, but still allowing execution and reading? No matter how we implement it, I think we want to make sure that any modes that re-use the container have restrictions preventing them from writing anywhere on the file system.

jakepruitt:
After chatting today, someone brought up the `--read-only` docker option (refs https://docs.docker.com/engine/reference/commandline/run/#mount-volume--v---read-only and `ReadOnlyRootFilesystem` https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-containerdefinitions.html)

This does exactly what this PR does manually. We should maybe try to do that instead.

jakepruitt:
Alright, after testing this out on the read-only file system it looks good for a test stack. Going to try on conex.

jakepruitt:
Successfully got the container recycling set up with ecs-conex at https://github.com/mapbox/ecs-conex/tree/watchbot4!

I think the approach here is nice and succinct. Going to merge it into the #184 PR now!

PR #212, opened 2018-06-09
Adding Reduce mode to ♻that container

tapasweni-pathak:
- [x] Closes https://github.com/mapbox/ecs-watchbot/issues/191.

Gonna hard and special test this PR; will be posting updates on the linked ticket.

Testing Plan.

- Outlined something in linked ticket would appreciate if someone from @mapbox/platform-engine-room has thoughts or better understanding. 
- Dependency: https://github.com/mapbox/watchbot-progress.
  -  watchbot-progress supports Javascript and Bash.
  1. Update this ticket with small writeup about how exactly Reduce Mode works with watchbot and watchbot-progress.
  2. Create a Reduce Job in Javascript see if it works watchbot-progress master.
  3. Create a Reduce Job in Bash see if it works with watchbot-progess master.
  4. Quite a few services don't use ecs-watchbot reduce mode and have set it up for themselves. I think deploying a existing running Map reduce service is easy, going to try with building reduce job with scripts. 
  5. Update watchbot-reduce examples/docs in watchbot-progress.

Please find more information on the linked ticket.  Gonna pick this up on Monday.

PR #221, opened 2018-06-18
Reduce mode 

tapasweni-pathak:
Branch for https://github.com/mapbox/ecs-watchbot/pull/212 as `git rebase master` is....

Updates in the linked PR.

cc @mapbox/platform-engine-room.

For some reason jest is not passing, I deleted the repo, cloned it and then again run `npn run update-jest-snapshots`, doesn't work. Tried with deleted package-lock.json and `rm -rf node_modules/` and then again `npm ci`, this has worked before.


arunasank:
> For some reason jest is not passing, I deleted the repo, cloned it and then again run npn run update-jest-snapshots, doesn't work. Tried with deleted package-lock.json and rm -rf node_modules/ and then again npm ci, this has worked before.

👋 @tapasweni-pathak, it seems like some of the jest dependencies are malfunctioning. To fix:
* Reset your version of package-lock.json to what's on master (So, revert the changes to package-lock.json alone on https://github.com/mapbox/ecs-watchbot/pull/221/commits/9f219d9659e416a49f760cc3f3970bdca86eb864)
* Run `npm run update-jest-snapshots`

I have tested that this works ^, and is the root cause. Let me know if it works for you! 🙂

tapasweni-pathak:
👋@arunasank I have to delete package-lock.json and re run npm ci once as I added a package, can't keep the master one. Plus as I wrote I delete the directory, reclone and install and run update-jest-snapshots it still doesn't work.

arunasank:
> 👋@arunasank I have to delete package-lock.json and re run npm ci once as I added a package, can't keep the master one. Plus as I wrote I delete the directory, reclone and install and run update-jest-snapshots it still doesn't work.

The reason the jest portion is not working is because when you're deleting the package-lock.json and running `npm install`, the dependencies for jest are getting updated (along with the dependencies for the new packages you're interested in installing), and these updated dependencies seem to be malfunctioning on your branch and others. Since you need to delete package-lock.json for using your new packages, maybe you can:
* Use the package-lock.json from master to update the jest snapshots
* Then delete package-lock.json and install your new modules.

While this will fix the issue of updating the snapshots, tests will still fail on Travis, because of the updated `package-lock.json`. Going to check if there's a way to partially update the `package-lock.json` file.

tapasweni-pathak:
@arunasank We can voice on it, this is weird,

1. `rm -rf ecs-watchbot/`.
2. `git clone git@github.com:mapbox/ecs-watchbot.git`.
3. `cd ecs-watchbot`
4. `git fetch origin`
5. `rm package-lock.json`
7. `npm install && npm link` (was using npm ci)
8.  `npm run update-jest-snapshots`


This is when I need to update package-lock.json. I used the updated package-lock.json and it passed. I plan to ticket in jest, just curious what exactly is happening. I have had this with jest before as well. I agree dependecies updating, thanks for finding the time to write your notes.

arunasank:
> This is when I need to update package-lock.json. I used the updated package-lock.json and it passed. 

👍 Glad it's fixed!

.eslintrc
```diff
@@ -6,7 +6,7 @@
     "jest": true
   },
   "parserOptions": {
-    "ecmaVersion": "2017"
+    "ecmaVersion": 8
```
jakepruitt: Why the switch to ecmaVersion 8 here?

lib/template.js
```diff
@@ -145,6 +155,48 @@ module.exports = (options = {}) => {
...
+  const ref = {
+    logGroup: cf.ref(prefixed('LogGroup')),
+    topic: cf.ref(prefixed('Topic')),
+    queueUrl: cf.ref(prefixed('Queue')),
+    queueArn: cf.getAtt(prefixed('Queue'), 'Arn'),
+    queueName: cf.getAtt(prefixed('Queue'), 'QueueName'),
+    notificationTopic: cf.ref(prefixed('NotificationTopic'))
+  };
+
+  const reduce = (prefixed, Resources, options, ref) => {
```
jakepruitt: What is your goal with making this section a function, rather than putting this code inside the `if` statement on line 237?

.eslintrc
tapasweni-pathak: I think ecmaVersion released in 2017 is 8th edition, so can be written as 2017 or 8.

lib/template.js
tapasweni-pathak: Can change it to be inside the `if` statement on line 237, added it to be consistent with the other methods defined in the template like `mount` e.t.c.

bin/watchbot-log.js
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env node
+'use strict';
+
+/**
+ * watchbot-log "something that you want logged"
+ *   - or -
+ * echo "somehing that you want logged" | watchbot-log
```
jakepruitt: We've removed watchbot-log from watchbot 4. Could you delete this file?

lib/template.js
```diff
@@ -145,6 +155,39 @@ module.exports = (options = {}) => {
     );
   };
 
+	const reduce = (prefixed, Resources, options, ref) => {
```
jakepruitt: I still think the fact that this function is only used once and modifies `Resources`, `options` and `ref`, we shouldn't have the code live in a function, but instead have it live inside of the `if` statement at line 227. If we do want to keep it as a function, we should have that function live right next to the if statement at line 227 so we can easily reference the function code.

bin/watchbot-log.js
tapasweni-pathak: Yeah, did a `git pull origin master` which didn't pull this change for some reason. Not checking nits at the moment. Thanks for pointing it out, will update.

lib/template.js
tapasweni-pathak: Nice, will do, added it for consistency and correctness like other function definitions.

tapasweni-pathak: Nice, will do, added it for consistency and correctness like other function definitions.

jakepruitt:
This looks good!

Issue #216, opened 2018-06-14
Metrics dashboard in watchbot 4


jakepruitt:
We should create a dashboard in the cloudformation template for watchbot stacks that comes with a few graphs:

- CPUUtilization and MemoryUtilization
- DesiredCapacity and RunningCapacity
- ApproximateMessagesNotVisible and ApproximateMessagesVisible
- NumberOfMessagesDeleted
- Is there a logs widget? Or anything we'd want to put in a text widget?

cc/ @mapbox/platform

arunasank:
> * Is there a logs widget? Or anything we'd want to put in a text widget?

There is no logs widget. The text-widget is if we wanted to add something in markdown, which we don't really need. First draft of the dashboard is up!

PR #222, opened 2018-06-18
Set up a metrics dashboard

arunasank:
Fixes https://github.com/mapbox/ecs-watchbot/issues/216

## TODO
* [x] Test this out, to see what the dashboard looks like


arunasank:
What the dashboard looks like:

screen_shot_2018-06-18_at_9_31_27_pm

lib/dashboard.js
```diff
@@ -0,0 +1,105 @@
...
+    {
+      type: 'metric',
+      x: 0,
+      y: 0,
+      width: 12,
+      height: 6,
+      properties: {
+        view: 'timeSeries',
+        stacked: false,
+        title: 'WatchbotQueue: ApproximateNumberOfMessagesNotVisible, ApproximateNumberOfMessagesVisible',
```
jakepruitt: Nitpick: could we make this name shorter? I don't think the full title would fit in the dashboard UI. Could we maybe change it to `Visible and NotVisible Messages`?

lib/dashboard.js
```diff
@@ -0,0 +1,105 @@
...
+      x: 0,
+      y: 12,
+      width: 12,
+      height: 6,
+      properties: {
+        view: 'timeSeries',
+        stacked: false,
+        title: 'WatchbotQueue: NumberOfMessagesDeleted',
+        metrics: [
+          ['AWS/SQS', 'NumberOfMessagesDeleted', 'QueueName', '${AWS::StackName}-WatchbotQueue', { period: 60 }]
```
jakepruitt: Is there a place to specify the `Statistic` to assess for this metric? I believe we want the `Sum` here, and the screenshot you made was using the `Average`.

lib/dashboard.js
```diff
@@ -0,0 +1,105 @@
...
+      width: 12,
+      height: 6,
+      properties: {
+        view: 'timeSeries',
+        stacked: false,
+        title: 'Scaling as a function of the rate of processing',
+        metrics: [
+          ['Mapbox/ecs-cluster', 'RunningCapacity', 'ClusterName', '${Cluster}', 'ServiceName', '${WatchbotService}', { period: 60 }],
+          ['.', 'DesiredCapacity', '.', '.', '.', '.', { period: 60 }],
+          ['AWS/SQS', 'NumberOfMessagesDeleted', 'QueueName', '${AWS::StackName}-WatchbotQueue', { period: 60 }]
```
jakepruitt: Again, we'll want this metric to be `Sum`.

lib/dashboard.js
```diff
@@ -0,0 +1,105 @@
...
+    {
+      type: 'metric',
+      x: 12,
+      y: 12,
+      width: 12,
+      height: 6,
+      properties: {
+        view: 'timeSeries',
+        stacked: false,
+        title: 'Scaling as a function of the rate of processing',
```
jakepruitt: Since scaling isn't triggered by the rate of processing, I think the use of `as a function of` here is misleading. Could we change the name of this title to something like `Concurrency vs. Throughput`, or possibly `Capacity vs. Progress`

Taking a step back, do we need this graph? We have the capacity on the graph above, and the `NumberOfMessagesDeleted` to the left of this graph. I feel like this graph may confuse people.

arunasank: imho this graph is the most useful, since this is the one most of us used for understanding scaling behaviour, and I feel pretty strongly about keeping it for that reason to spare folks the effort and time of generating it/figuring out what metrics are most relevant to generate this.

👍 on the name change and the statistic comments. I need to do a final check of the names and statistics - will fix these.

arunasank:
@jakepruitt I also want to add some DeadLetterQueue metrics, specifically the `ApproximateNumberOfMessagesVisible` and `ApproximateNumberOfMessagesNotVisible`! Will do that.

Issue #218, opened 2018-06-16
Where'd the docs go?


rclark:
A whole lotta docs vanished with the merge of #184. We should bring them back ASAP.

cc @mapbox/platform-engine-room

jakepruitt:
Closed with #224

Issue #202, opened 2018-05-24
Installing watchbot 4


jakepruitt:
## Background

In order for users to run a watchbot 4 stack, they will need to install watchbot into their container’s environment in a way that exposes the `watchbot listen` command. This command will be the process run by the container at upstart to poll the queue, calling out to a subprocess with the user-provided command to process jobs.

The challenge is then figuring out how to install the `watchbot` command. While it’s probably easiest to use `npm install -g` `@``mapbox/watchbot`, we do not necessarily know if the container will have node/npm installed, or if that version of node/npm will be compatible with the async/await code inside of watchbot, which requires node version 8 or higher.

For that reason, we need to create an easier installation flow for users.

From @rclark in https://github.com/mapbox/ecs-watchbot/pull/184

> Since developers are going to need to "install" watchbot in their Dockerfiles, we should make that trivial, and it should bundle node v8 in a way that will allow watchbot to run independently of whatever node version the actual processing code requires.

## Ideal interface for installing watchbot 4

Here’s a rough sketch of how I imagine watchbot 4 could be installed in users’ Dockerfiles:

- When the Dockerfile is built, watchbot is downloaded via a `curl` command.
- The installation is done by a bash script, which sets up `watchbot` as an executable on the system, but does not override the default version of node.js on the system.
- After the Dockerfile is built and the user is ready to deploy the service, the cloudformation template created by `watchbot.template()` sets the `command` property in the task definition to be `command: 'watchbot listen <user-command>'`
- When `watchbot listen` is executed in the container, it implicitly uses node.js v8 to run `bin/watchbot.js`.
- When watchbot uses `child_process.spawn()` to create a worker, that worker has access to all of the environment packages set up in the Dockerfile, including whatever version of node.js the developer installed in the container, not the version installed by watchbot.

## Next actions

- Determine the URL where watchbot will be hosted. I imagine a github-based URL would be ideal here for open-source users.
- Write the bash script to install node.js version 8 in a place where it will not conflict with an already installed version of node.js. The installation script will also set the `watchbot` command up in the container’s `PATH` in a way that works on all linux operating systems.

cc/ @mapbox/platform-engine-room @emilymcafee

jakepruitt:
> Watchbot is downloaded via a curl command.

Adding a **bonus challenge** here - how do we implement the installation with confidence that the URL contents won't be overridden? Should the URL come from a protected branch on github? How secure are github release versions?

cc/ @mapbox/security

[...]

PR #224, opened 2018-06-18
Docs 4 watchbot 4

jakepruitt:
Started taking a stab at #218. @mapbox/platform-engine-room feel free to commit directly to this branch if you have more things to add.

### DIFF
--- a/docs/building-a-template.md
+++ b/docs/building-a-template.md
@@ -57,7 +57,6 @@ var watch = watchbot.template({
 module.exports = cloudfriend.merge(myTemplate, watch);
 ```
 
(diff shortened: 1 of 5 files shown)


arunasank:
Noting that we should document that logging has become much better now, and no longer requires services to use the `watchbot-log` binary. 

Ref: https://github.com/mapbox/ecs-watchbot/issues/202#issuecomment-398803880

arunasank:
@jakepruitt made some changes and additions to the docs, but on another PR: https://github.com/mapbox/ecs-watchbot/pull/229

PR #225, opened 2018-06-20
Logging in Watchbot4

### LINKED ISSUE #204: Logging in Watchbot4
### Generic TODO

- deciding on `FilterPattern` for correctly parsing logs for `MetricNamespace: 'Mapbox/ecs-watchbot'`, custom CloudWatch metrics
- logging corrections/enhacements
- any other logging which needs to be added
- What all `MetricNamespace: 'Mapbox/ecs-watchbot'`, CloudWatch metrics should be added to Watchbot4, `master` has
    - WorkerErrorMetric (has an alarm as WorkerErrorsAlarm)
    - WorkerDurationMetric
    - WorkerPendingMetric
    - WatcherConcurrencyMetric
    - FailedWorkerPlacementMetric (has an alarm as FailedWorkerPlacementAlarm) 

### Watchbot Master branch (current version)

For WorkerErrorMetric and FailedWorkerPlacementMetric, we have setup logging as
https://github.com/mapbox/ecs-watchbot/blob/master/lib/tasks.js#L99-L102

```js
tasks.report = function(outcome) {
  switch(outcome) {
    case tasks.outcome.success: return 'success';
    case tasks.outcome.noop: return 'no-op, will retry';
    case tasks.outcome.fail: return 'failed, removed from queue';
    case tasks.outcome.retry: return 'failed, will retry';
  }
};
```

and

```js
log.warn(
    '[%s] %s',
    env.MessageId,
    JSON.stringify({
        failedPlacement: 'true',
         reason: err.message
    })
);
```

The `FilterPattern` we set is

-  `{ $.outcome = "failed*" }` on [WorkerErrorsMetric](https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L316-L327)
-  `'{ $.failedPlacement = true }'`  on [FailedWorkerPlacementMetric](https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L288-L299).
- `'{ $.duration = * }',` on [WorkerDurationMetric](https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L344-L357)
- `'{ $.pending = * }',` on [WorkerPendingMetric](https://github.com/mapbox/ecs-watchbot/blob/master/lib/template.js#L359-L372)

In the watchbot master branch, we are able to have datapoints on custom metrics as we log this way.

### Watchbot 4(Watchbot container-recycling branch)
1. None of the above filter pattens would work as currently the logging setup is different.
2. logs prefixed with `[watcher]` : [ref](https://github.com/mapbox/ecs-watchbot/blob/container-recycling/lib/worker.js#L32), where they could be `[worker]` : [ref](https://github.com/mapbox/ecs-watchbot/blob/master/lib/tasks.js#L99-L102)
3. logs prefixed with `[failure]` has no special object like previous version, they would output something like, [ref](https://github.com/mapbox/ecs-watchbot/blob/container-recycling/lib/logger.js#L34).
```json
    [watcher] [<id>] [failure]
    {
        "code": 255,
        "signal": null,
        "duration": 7
    }
```
4.  A worker failing would log as
  -  `[error] [worker] ${err.message}`
  -  `[failure] ${JSON.stringify(results)}`

5. Queue error would log as 
  -  `[error] [sqs] ${err.message}`

6. Duration would be logged through [ref](https://github.com/mapbox/ecs-watchbot/blob/container-recycling/lib/worker.js#L15-L20)
```js
     const duration = Date.now() - start;
     resolve({ code, signal, duration });
  });
    
   child.stdout.pipe(logger.stream());
   child.stderr.pipe(logger.stream());

```

### Questions/Next steps
I’m thinking of

- WorkerErrors would be filtered with a metricFilter on `[error] [worker]`
- QueueError metric filtered on `[error] [sqs]`
- Duration metric filtered on `duration` value of object logged.
- Failed worker metric filtered on `[failure]`
- FailerWorkerPlacementMetric logs currently are only in service history.
- Create an object like `outcome` while logging in current version so that the logs can be correctly parsed with `{ $.outcome = "failed*" }` otherwise with container-recycling branch the metric filter would be set to `"[failure]"`.

Other than this I would enhance the logs to put `[worker]` , `[watcher]` when they are worker or watcher logs, which would then be added in `MetricFilter` value.

To make the logs parsable and parse them correctly for custom namespace CloudWatch metric does this cleanup/enhancement and filtering of logging looks good? Is there any other metric or logging prefixing or logging messages needs to be added which I'm missing here?

cc @rclark @mapbox/platform-engine-room.

tapasweni-pathak:
Closes #204.

cc @mapbox/platform-engine-room.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -31,11 +31,11 @@ class Logger {
   }
 
   workerFailure(results) {
-    this.log(`[failure] ${JSON.stringify(results)}`);
+    this.log(`[watchbot] [failure] ${JSON.stringify(results)}`);
   }
 
   workerError(err) {
-    this.log(`[error] [worker] ${err.message}`);
+    this.log(`[error] [watchbot] ${err.message}`);
   }
 
   queueError(err) {
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -31,11 +31,11 @@ class Logger {
   }
 
   workerFailure(results) {
-    this.log(`[watchbot] [failure] ${JSON.stringify(results)}`);
+    this.log(`[failure] ${JSON.stringify(results)}`);
   }
 
   workerError(err) {
-    this.log(`[error] [watchbot] ${err.message}`);
+    this.log(`[error] [worker] ${err.message}`);
   }
 
   queueError(err) {
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -45,26 +45,27 @@ class Worker {
     this.volumes = options.volumes;
     this.maxJobDuration = options.maxJobDuration;
     this.message = message;
-    this.logger = Logger.create('watcher', message);
+    this.watcherLogger = Logger.create('watcher', message);
+    this.workerLogger = Logger.create('worker', message);
   }
 
   async success(results) {
-    this.logger.workerSuccess(results);
+    this.watcherLogger.workerSuccess(results);
     return await this.message.complete();
   }
 
   async ignore(results) {
-    this.logger.workerSuccess(results);
+    this.watcherLogger.workerSuccess(results);
     return await this.message.complete();
   }
 
   async noop(results) {
-    this.logger.workerSuccess(results);
+    this.watcherLogger.workerSuccess(results);
     return await this.message.retry();
   }
 
   async fail(results) {
-    this.logger.workerFailure(results);
+    this.watcherLogger.workerFailure(results);
     return await this.message.retry();
   }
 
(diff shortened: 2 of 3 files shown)


jakepruitt:
Per conversation, we will need to create two different loggers inside of the `lib/worker.js`:

```
constructor(message = {}, options = {}) {
  // ...

  this.workerLogger = Logger.create('worker', message);
  this.watcherLogger = Logger.create('watchbot', message);
}
```

The `workerLogger` will be used within the `child` method as the destination of the `child.stdout.pipe(workerLogger.stream())`, that way all child logs will be prefixed with `worker`.

The `watcherLogger` will be used everywhere else in the file, like the `success` and `failure` messages that are meta information surrounding the child process.

Issue #213, opened 2018-06-10
Use the stack name in the containerName


arunasank:
Currently, the `containerName` uses the `options.service` string, which is configured by the watchbot client. Since the string is not created using CloudFormation helper functions, the `containerName` cannot be customised based on the stack name. This means that if multiple stacks are running on the same service, the containerName reflects the service name and not the stack name and this can be confusing if you're parsing `Cloudtrail` logs and looking for containers of a specific stack.

To fix, we can change https://github.com/mapbox/ecs-watchbot/blob/c950b97be2864f4027faa77d098ffd06cd4ded61/lib/template.js#L667 to concatenate using cloudfriend helper functions like `Fn::Join`

cc/ @mapbox/platform-engine-room

PR #226, opened 2018-06-20
Use the stack-name in the `Name` property of the ContainerDefinition

arunasank:
Fixes https://github.com/mapbox/ecs-watchbot/issues/213

Currently, the `options.service` variable is required to be a string since it's concatenated with the `options.prefix` to create the `Name` property of the container. This PR removes that requirement and allows the `options.service` to be a Cloudformation reference. Besides the flexibility that this allowance gives us, it is also useful for differentiating containers of different stacks of the same repository, since the `options.service` can be referenced to `AWS::StackName` now, for example.

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -344,7 +344,7 @@ module.exports = (options = {}) => {
       Family: options.family,
       ContainerDefinitions: [
         {
-          Name: cf.join('-', [prefixed(''), options.service]),
+          Name: cf.join('-', [prefixed(''), cf.stackName]),
           Image: cf.join([
             cf.accountId,
             '.dkr.ecr.us-east-1.amazonaws.com/',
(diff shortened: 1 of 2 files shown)


arunasank:
Decided to remove the `options.service` dependency altogether from the `Name` property of the container, since the `options.service` property if also used to derive the name of the ECR repository for a stack. Using the `cf.stackName` reference directly in the `Name` property, to avoid clashing with the `options.service`.

arunasank:
Additional tests are not required, since the changes use a cloudformation reference in the container name (and don't modify the datatype of the options.service variable as intended earlier), and the fact that the existing tests in `template.validation.test.js` pass is enough proof that the changes don't introduce validation errors in the cloudformation template. I've also tested this on a watchbot stack and double-confirmed that there are no validation errors, and that the name of the container is as expected.

arunasank:

PR #227, opened 2018-06-20
Remove watchbot-log binary

arunasank:
Quick PR to have the watchbot binary perform both `listen` and `log` tasks. We can deprecate the watchbot-log binary since it is not backward-compatible anyway, owing to the Node 7.6+ dependency of the `Logger` class.


arunasank:
Per https://github.com/mapbox/ecs-watchbot/issues/202#issuecomment-398803880, I've deleted the watchbot-log binary from watchbot (4) 4ever! We should record this deprecation in our docs.

PR #236, opened 2018-06-27
Only expose template, which can run on node 6

jakepruitt:
Exposes the template, but removes the exposed modules that rely on node 8. People can now locally run node 6 and compile the template.


jakepruitt:
Confirmed this works on https://github.com/mapbox/ecs-telephone! Going to merge and release as a patch.

PR #237, opened 2018-06-27
Remove engine requirement

jakepruitt:
Removes the node 8 engine requirement.


rclark:
Hmm. Could you say why this is being removed?

jakepruitt:
@rclark the engine requirement throws errors for downstream users that want to use node version < 8, even when they specify the dependency as a dev dependency.

Ideally, there isn't a node version requirement for watchbot users.

PR #240, opened 2018-07-02
Allow empty sns Subject

tcql:
Currently, If an SNS message is sent with no Subject parameter, the `$Subject` in the worker script is set to the literal word "undefined". This could be breaking for stacks that respond to multiple Subject values that expect an empty/default subject value to be an unset variable; this was the old behaviour

cc @jakepruitt @rclark

### DIFF
--- a/lib/message.js
+++ b/lib/message.js
@@ -36,7 +36,6 @@ class Message {
 
     this.env = {
       MessageId: sqsMessage.MessageId,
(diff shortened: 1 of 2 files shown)


tcql:
hah was already on that @jakepruitt but you reviewed faster than i could write the test! you're fast ⚡️

PR #243, opened 2018-07-03
Update upgrade install process, recommend binary install over NPM

k-mahoney:
Adds distinct install steps depending on Node.js version.

/cc @mapbox/platform-engine-room

### DIFF
--- a/docs/upgrading-to-watchbot4.md
+++ b/docs/upgrading-to-watchbot4.md
@@ -8,17 +8,14 @@
 ## Steps:
 - Move the CMD in Dockerfile to the `command` option in the cloudformation template
 - Install watchbot’s binary with `wget` in your Dockerfile:
+   ```
+   # The `<os>` can be `linux`, `macosx`, `windows`, depending on what you're running on your container.
+   # The `<tag>` is a tag on https://github.com/mapbox/ecs-watchbot/tree/master that is more recent than v4.0.0
 
+   RUN wget https://s3.amazonaws.com/watchbot-binaries/<os>/<tag>/watchbot -O /usr/local/bin/watchbot
+   RUN chmod +x /usr/local/bin/watchbot
 ```
-# The `<os>` can be `linux`, `macosx`, `windows`, depending on what you're running on your container.
-# The `<tag>` is a tag on https://github.com/mapbox/ecs-watchbot/tree/master that is more recent than v4.0.0
-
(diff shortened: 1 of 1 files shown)


docs/upgrading-to-watchbot4.md
```diff
@@ -7,13 +7,13 @@
 
 ## Steps:
 - Move the CMD in Dockerfile to the `command` option in the cloudformation template
-- Install watchbot’s binary with `wget` in your Dockerfile:
-
-```
-    RUN wget https://s3.amazonaws.com/mapbox/watchbot/linux/watchbot -O /usr/local/bin/watchbot
-    RUN chmod +x /usr/local/bin/watchbot
-```
+- If you are running a Node.js version older than `8.x`, install watchbot’s binary with `wget` in your Dockerfile:
```
jakepruitt: jake opinion: I think `version less than 8.x` might make more sense here than `version older than 8.x`. The fact that "older" means "smaller number" kinda breaks my brain. Where does your moral compass point on this one @k-mahoney?

k-mahoney: Yeah, that makes more sense - will update 👍

rclark:
My only question is: should we just say that you install via `wget` **always**? What's the benefit of advertising more than one installation pathway?

k-mahoney:
In reviewing, most stacks include `@mapbox/watchbot` in their `package.json` and install it alongside other dependencies via `npm install` in the Dockerfile. If we want to advise the binary file over this method that's fine, as it removes the necessity for Node version compatibility.

I assume there is no specific harm if a stack installs both the binary file and `@mapbox/watchbot` using `npm`? Beyond a stack that (1) installs the binary, (2) retains watchbot in its `package.json`, and (3) runs an incompatible version of Node - resulting in `npm install` failing?

rclark:
> most stacks include @mapbox/watchbot in their package.json and install it alongside other dependencies via npm install in the Dockerfile

We may actually get to a better place if we actively remove this behavior. While right now that `npm install` step may fail because the application is written for node.js 4 and watchbot needs 8, there's a future where it fails for the opposite reason: the application has upgraded to node.js 14 and watchbot is still using 8.

If we rally around the `wget` installation, we have a more clear separation of concerns moving forward.

k-mahoney:
Okay, that sounds good to me @rclark. Recommended steps here could be:
1. List `@mapbox/watchbot` under `devDependencies` rather than `dependencies`
2. If required, `npm install` in a Dockerfile should use `npm install --production`
3. Watchbot should be installed via the binary only

PR #248, opened 2018-07-05
Fix command for jest snapshot updation

arunasank:
Ref: https://github.com/mapbox/ecs-watchbot/pull/220#issuecomment-402598855


jakepruitt:
Whoops, didn’t know about this command, thanks for the hot fix @arunasank!

PR #249, opened 2018-07-05
custom cloudformation resource for scale up adustment

brendanmcfarland:
Goals here: 

- Allow `maxSize` to be parameterized in templates
- scaleUp by approximately 10% of the max size (round to an integer)
  - Max scaling adjustment of 100
  - Min scaling adjustment of 1

Tested this out with maxSize as a cf.ref and updating maxSize to 3333, 33, 250, etc. Looked 🆗 to me.

cc: @jakepruitt @arunasank @k-mahoney @rclark

### DIFF
--- a/lib/template.js
+++ b/lib/template.js
@@ -702,7 +702,7 @@ module.exports = (options = {}) => {
         ZipFile: cf.sub(`
           var response = require('cfn-response');
           exports.handler = function(event,context){
-            var result = Math.min(parseInt(event.ResourceProperties.maxSize) / 10, 100)
+            var result = parseInt(Math.max(Math.min(parseInt(event.ResourceProperties.maxSize) / 10, 100), 1));
             response.send(event, context, response.SUCCESS, { ScalingAdjustment: result });
           }
           `)
--- a/lib/template.js
+++ b/lib/template.js
@@ -470,7 +470,7 @@ module.exports = (options = {}) => {
         MetricAggregationType: 'Average',
         StepAdjustments: [
           {
-            ScalingAdjustment: Math.ceil(options.maxSize / 10),
+            ScalingAdjustment: cf.getAtt(prefixed('CustomScalingResource'), 'ScalingAdjustment'),
             MetricIntervalLowerBound: 0.0
           }
         ]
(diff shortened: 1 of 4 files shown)


brendanmcfarland:
Yep, when passing in garbage the response is:

`ERROR: Failed to create stack: ValidationError: Parameter 'maxSize' must be a number.`

Or:

`ERROR: Failed to generate changeset: ValidationError: Parameter 'maxSize' must be a number.`

jakepruitt:
Nice!

arunasank:
Awesome! 🙌

brendanmcfarland:
Ah - I've realized that you could send a string parameter in if you changed the type of your cloudormation parameter from `Number` to something else. In this case the failure hits after the validation and during the deploy at:

> WatchbotScalingTarget: Encountered non numeric value for property MaxCapacity

I'm against error handling in the custom resource; because if that lambda function errors the stack ends up in an hour long wait before ending in an UpdateRollbackFailed state. 

I don't see that we've done any validation on other fields like CPU, minSize, etc. So it feels _okay_ to me to leave this as the way a user encounters failure if they put in a non-numeric maxSize.

lib/template.js
```diff
@@ -664,6 +664,63 @@ module.exports = (options = {}) => {
...
+  },
+
+  Resources[prefixed('ScalingLambda')] = {
+    Type: 'AWS::Lambda::Function',
+    Properties: {
+      Handler: 'index.handler',
+      Role: cf.getAtt(prefixed('ScalingRole'), 'Arn'),
+      Code: {
+        ZipFile: cf.sub(`
+          var response = require('cfn-response');
```
rclark: oooooh good find.

lib/template.js
```diff
@@ -664,6 +664,63 @@ module.exports = (options = {}) => {
...
+  },
+
+  Resources[prefixed('ScalingLambda')] = {
+    Type: 'AWS::Lambda::Function',
+    Properties: {
+      Handler: 'index.handler',
+      Role: cf.getAtt(prefixed('LambdaScalingRole'), 'Arn'),
+      Code: {
+        ZipFile: cf.sub(`
+          var response = require('cfn-response');
```
jakepruitt: _nitpick_ `var` => `const`?

PR #250, opened 2018-07-06
bind this reference

brendanmcfarland:
Looking to address an error we've seen on the heartbeat:

```
(node:1) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'queueError' of undefined
    at Timeout.heartbeat [as _onTimeout] (/snapshot/ecs-watchbot/lib/message.js:82:19)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5)
```

Looks related to how `setInterval` handles `this`

cc: @jakepruitt

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -80,7 +80,7 @@ class Worker {
       stdio: [process.stdin, 'pipe', 'pipe']
     };
 
-    const heartbeatTimeout = setInterval(this.message.heartbeat, 120000);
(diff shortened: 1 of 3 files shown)


brendanmcfarland:
@jakepruitt mind giving this a quick review - was able to replicate the error above with master, and this fix seems to have eliminated the undefined `this`. Worked with the initial message a worker processed and each of the subsequent ones.

Issue #258, opened 2018-07-10
Cannot read property `1` error in /bin/watchbot-binary-generator


brendanmcfarland:
Seeing errors in codepipeline:

```
(node:73) UnhandledPromiseRejectionWarning: TypeError: Cannot read property '1' of null
    at data.forEach (/codebuild/output/src598842203/src/bin/watchbot-binary-generator.js:25:43)
    at Array.forEach (<anonymous>)
    at Promise (/codebuild/output/src598842203/src/bin/watchbot-binary-generator.js:21:10)
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:228:7)
(node:73) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:73) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Appears to be happening on the commits that are tagged. Not sure what changed that is breaking this script.

https://github.com/mapbox/ecs-watchbot/blob/0a9b16a2e200b136d934c90d87d098ef8592df86/bin/watchbot-binary-generator.js#L17-L29

cc: @arunasank @jakepruitt

PR #256, opened 2018-07-10
docs/development: initial version

freenerd:
While working on https://github.com/mapbox/ecs-watchbot/pull/251 I realized that there are no releasing docs. So I wrote some.

Todo:

- [ ] 👀@jakepruitt 
- [ ] Improvements
- [ ] Merge

PR #259, opened 2018-07-10
Trigger binary creation when tags are created manually

arunasank:
Fixes https://github.com/mapbox/ecs-watchbot/issues/258

I created a fresh "prerelease" version (4.7.2-0), and saw that the build was successful. To reproduce the error:
* I set the `sha` variable on https://github.com/mapbox/ecs-watchbot/blob/master/bin/watchbot-binary-generator.js#L17 to `4506f753e0955ee3683e8c396cdeea58e4442c5e`
* This caused the same error to be thrown 

On adding more debugging, I found that the error is thrown when we create tags manually using `git tag -m ....` because these tags look like `4.7.0`, whereas the ones created using the `npm version...` command look like `v4.7.0`. 

I've modified the regex to allow numeric tags that are created manually as well, but strongly recommend creating tags using `npm version <patch|minor|major|prerelease>`.

Thanks for creating https://github.com/mapbox/ecs-watchbot/pull/256/files @freenerd  - it should help clarify a lot of this. 🙂 

### DIFF
--- a/bin/watchbot-binary-generator.js
+++ b/bin/watchbot-binary-generator.js
@@ -21,7 +21,7 @@ const getTagForSha = async (sha) => {
     data.forEach((ref) => {
       ref = ref.split('\t');
       if (ref[0] !== sha) return;
-      const tagRegex = /refs\/tags\/(v[0-9.-]+)(\^\{(.*)\})*/;
+      const tagRegex = /refs\/tags\/(v?[0-9.-]+)(\^\{(.*)\})*/;
       return resolve(tagRegex.exec(ref[1])[1]);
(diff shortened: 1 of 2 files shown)


jakepruitt:
For consistency, should we add a `v` if it's missing before writing it to S3?

arunasank:
> For consistency, should we add a v if it's missing before writing it to S3?

I think this would be confusing, since the path is conceived using the tags present on GitHub, and if we are changing the naming of these tags, we're modifying it to be different from the actual GitHub tag. I think we should use the exact same tag string on GitHub.

PR #266, opened 2018-07-12
Build binary for alpine

rclark:
Builds a binary for alpine linux. Also just some slight cleanups in the binary-building script.


arunasank:
Thanks for merging this in @rclark. Also nice branch name. 😄

PR #267, opened 2018-07-13
Combo metric

jakepruitt:
Combines the `MessagesVisible` and `MessagesNotVisible` metrics into a minutely `TotalMessages` metric that we should scale on.

cc/ @rclark @brendanmcfarland


brendanmcfarland:
Shoot I knew there was a reason I didn't use node 8 in the custom resource inline lambda...

> ZipFile can only be used when Runtime is set to either of nodejs, nodejs4.3, nodejs6.10, python2.7, python3.6

@jakepruitt can remove the node8 `await`s etc in the code and use 6.10, or use a different code param like an S3 bucket/key

rclark:
> ZipFile can only be used when Runtime is set to either of nodejs, nodejs4.3, nodejs6.10, python2.7, python3.6

This makes me nauseous

Issue #255, opened 2018-07-10
Docs to explain how watchbot binaries are generated


arunasank:
Let's add some docs to capture the entire process of how watchbot binaries are generated, how that stack is deployed to, etc.

cc/ @mapbox/platform-engine-room

PR #270, opened 2018-07-16
Docs on watchbot binaries

arunasank:
Fixes https://github.com/mapbox/ecs-watchbot/issues/255 when merged.

### DIFF
--- a/docs/watchbot-binaries.md
+++ b/docs/watchbot-binaries.md
@@ -1,0 +1,24 @@
+# Watchbot binaries
+
+## What is a watchbot binary?
+A watchbot binary contains all of the watchbot code and dependencies,  pre-packed using [`zeit/pkg`](https://github.com/zeit/pkg/) and node 8, for the linux, mac OSX, windows and Alpine operating systems.
+
+## Why do we need watchbot binaries?
+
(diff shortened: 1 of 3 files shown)


docs/watchbot-binaries.md
```diff
@@ -0,0 +1,28 @@
+# Watchbot binaries
+
+## What is a watchbot binary?
+A watchbot binary contains all of the watchbot code and dependencies,  pre-packed using [`zeit/pkg`](https://github.com/zeit/pkg/) and node 8, for the linux, mac OSX, windows and Alpine operating systems.
+
+## Why do we need watchbot binaries?
+
+The latest upgrade of `ecs-watchbot`, watchbot 4, also upgrades the node engine that watchbot uses. While this allows watchbot to use the latest JavaScript programming constructs, as well as the most optimised node engine, it also constrains consumers of watchbot to upgrade their node dependency if they are triggering a watchbot call with local watchbot scripts. As a workaround, we are pre-packaging watchbot, node 8 and associated scripts as a binary which helps consumers use the latest watchbot, without upgrading their existing node dependency.
```
rclark: The node version mismatch is really a transient problem. The real reason to include binaries is that watchbot is now an application that you have to explicitly add to your dockerfile. Its a standalone application and so we should provide a simple installation solution. The developer really shouldn't even have to know that watchbot uses node.js -- I would focus the discussion on making it easy to install and configure, rather than on node version mismatch.

arunasank:
> I'd love to see a big fat copiable line right at the top for Install the current watchbot version. This would have to be updated when new versions of watchbot are created but is immensely useful.

I moved this to the top of the readme file for more visibility and made the other fixes. Thanks for the review @rclark!

PR #275, opened 2018-08-07
Enforce 12 hour maxJobDuration

### LINKED ISSUE #203: The timeout problem in watchbot 4
## Background

With the distributed nature of watchbot 4’s polling, we resurface an old problem with receiving messages that are currently processing but have timed out their message timeouts and been sent back into the queue.

From @rclark:

> I just thought of another watchbot regression at v4: 
> 
> Right now watchbot is "smart" about detecting the situation where your SQS messages time out before they finish. The watcher cross-checks an incoming message with receive > 1 against SQS ids that it is currently processing, and if they match, that means its a "duplicate receive". It used to be that I told everyone to solve this by changing sqs message timeouts, which are configurable via watchbot, but confusing as hell to everyone.
> 
> At v4, you can't check for duplication because watchers are distributed.

Important reading on duplicate receives: https://github.com/mapbox/ecs-watchbot/issues/175

Further reading https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html

## Next actions

We should draft up a list of potential options to mitigate this situation. @rclark’s solution in watchbot 3 is outlined here: https://github.com/mapbox/ecs-watchbot/pull/176. SQS suggests a “heartbeat” system in https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html:

> If you don't know how long it takes to process a message, create a heartbeat for your consumer process: Specify the initial visibility timeout (for example, 2 minutes) and then—as long as your consumer still works on the message—keep extending the visibility timeout by 2 minutes every minute.

We also have some interesting methods of implementing hard timeouts on containers by killing sub-processes if we wanted to go in a more brutal direction.

The floor is open for other suggestions from @mapbox/platform-engine-room. Ideally we can get rolling on some implementations and tests soon.

cc/ @emilymcafee @rclark

brendanmcfarland:
Opening a pr to close the loop on timeout handling. 

- we can't extend message visibility past 12 hours 
- we don't have a single source of truth to avoid duplicate work

As it stands right now watchbot will duplicate work if the message takes longer than 12 hours to process and no timeout is set. Quick solution is to acknowledge the current 12 hour limitation by enforcing maxJobDuration on workers.

I've chosen 43020 as the max (12 hours minus three minutes) to avoid edge cases that could occur if the heartbeat tries to extend the visibility timeout 3 minutes and the total would be over 43200 seconds. Also attempted some docs to explain this limitation.

cc: @jakepruitt @rclark 

closes #203


lib/template.js
```diff
@@ -63,9 +63,10 @@ const dashboard = require(path.resolve(__dirname, 'dashboard.js'));
  * number or a reference, i.e. `{"Ref": "..."}`.
  * @param {Boolean} [options.privileged=false] - give the container elevated
  * privileges on the host container instance
- * @param {Number|ref} [options.maxJobDuration] - the maximum number of seconds
+ * @param {Number|ref} [options.maxJobDuration] - the number of seconds
```
jakepruitt: Should we say that the default is 12 hours?

docs/building-a-template.md
```diff
@@ -103,6 +103,14 @@ In writableFilesystem mode, the whole file system is writable and containers are
 
 writableFilesystem mode has no restrictions to the file system: workers can write anywhere and read from anywhere, their files being instantly deleted after the job finishes and the container dies.
 
+### maxJobDuration explained
+
+When maxJobDuration is exceeded by a worker, the worker will stop all processes and return the processing message to the queue. Typically, the maxJobDuration should be set marginally above the maximum time that it takes your application to process and delete a message from the queue. This ensures all work can be completed; while providing protection against erroneous processing continuing to the maximum of twelve hours.
```
KeithYJohnson: I'm likely confused but it looks like if `maxJobDuration` is exceeded, the behavior is the same as when SQS hard limit of 12 hours is exceeded, because in both cases the message is returned to the queue, is that right?  

Or will a worker keep working when the 12-hour visibility timeout is breached?

rclark: > Or will a worker keep working when the 12-hour visibility timeout is breached?

Yes, I believe the worker will keep working regardless of hitting the timeout. We could implement a hard cutoff, but that may or may not be desirable depending on the application context.

KeithYJohnson: Ah makes sense, thanks @rclark!

PR #279, opened 2018-08-21
FIFO queues

### LINKED ISSUE #278: FIFO queue
We'd like to use a FIFO SQS queue instead of the standard SQS queue that Watchbot currently ships with.

@rclark I know you started work on this in https://github.com/mapbox/ecs-watchbot/compare/fifo. Do you have any idea how far out this might be? And is there anything we could do to help it along?

In some discussion you suggested this FIFO branch "hard-wires you to performing builds one at a time". In theory, I *think* a FIFO queue should allow multiple simultaneous jobs as long as they have different message group IDs. It sounds like the downside of this is that the new scaling logic wouldn't work nicely. If you're expecting the queue to include only one message group, then you'd only ever need one machine and it would be wasteful to spin up *more* machines based on the length of the queue. However, if you're expecting the queue to include multiple message groups (as we are), then with only one machine you'd be slowing things down, requiring messages in unrelated groups to wait for each other.

One idea: What if the scaling behavior for a FIFO queue is configurable? You could decide whether you accept one build at a time (because you won't be using multiple message groups at all or very much), or you accept the additional costs of scaling (because you expect that you'll *rarely* have messages from the same group overlapping).

Another idea: Maybe I'm thinking about this wrong, and should instead make our own FIFO queue that triggers a CodeBuild project?

cc @mapbox/frontend-platform @jseppi @emilymdubois

rclark:
This PR explores adding a configuration option that allows you to launch a watchbot stack that operates against a FIFO queue. On the first pass, the biggest adjustments watchbot and its users need to be aware of:

- Usually watchbot makes an SNS topic to feed an SQS queue. When an SQS message is received, it is expected that the message came via SNS, which leads to the `Subject` and `Message` (SNS message attributes) being parsed and exposed to the worker as environment variables. You can't send messages to a FIFO queue via SNS topic -- instead you have to send them directly to the queue. This means the parsing assumptions that watchbot makes may be invalid. The least invasive approach I can think of here would be to adjust message parsing in a backwards compatible way, and expose the `MessageGroupId` to the worker as the `Subject` environment variable, and the `Message` as the SQS message body.

- When feeding a message to a FIFO queue, you must provide a `MessageGroupId`. This acts as a sort of "shard" for the queue -- FIFO behavior is only guaranteed for messages that share an identical `MessageGroupId`. We need some additional documentation in here to explain this.

- Scaling behavior in watchbot is based on the total number of messages in the queue. If there are 50 messages, concurrent processing scales up to 50 (or the configured `MaxSize` if lower). In a FIFO queue, really you would want to scale on the number of `MessageGroupId`s that are being fed to your queue. If there are 50 messages in the queue, but only 2 `MessageGroupId`s, then watchbot would launch 50 containers for processing, but only 2 at a time would be able to process messages.

Currently this PR locks the number of containers to `1` if you specify `fifo: true`, and turns off autoscaling entirely. This is probably not the right approach to take. Given that something like "number of MessageGroupIds" is not a metric that SQS provides in cloudwatch, and given that it is entirely dependent on how the user sends messages to the queue, I don't think we're going to find an ideal scaling behavior here. We should probably re-enable scaling, and document that in a FIFO queue, its important that you consider up-front how many `MessageGroupIds` your queue will have, and set your stack's `MaxSize` to that value or lower.

fixes #278 
cc @davidtheclark

### DIFF
--- a/docs/using-a-fifo-queue.md
+++ b/docs/using-a-fifo-queue.md
@@ -1,0 +1,27 @@
+## Using a first-in-first-out (FIFO) queue
+
+By default, Watchbot creates a standard SQS queue, in which the order of jobs is not guaranteed to match the order of messages. If your program requires more precise ordering, you can use [a first-in-first-out (FIFO) queue](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html), instead.
+
+To use a FIFO queue, set the template option `fifo: true`.
+
+A FIFO queue behaves differently and has some limitations, so you should read [the AWS documentation](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html) and the details below if you are considering the switch.
+
+### Triggering FIFO workers
+
+With a standard queue, you trigger a Watchbot worker by sending an SNS message to the watcher's topic. With a FIFO queue, you send a message directly to the watcher's FIFO queue, instead. Make sure you have read [the AWS documentation](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html) about the expected format of FIFO queue messages (for example, you need to include a `MessageGroupId`).
+
+### Scaling FIFO workers
+
+[Scaling behavior](./scaling-in-watchbot.md) in Watchbot is based on the total number of messages in the queue. If there are 50 messages, concurrent processing scales up to 50 (or the configured `maxSize`, if lower). In a FIFO queue, this scaling will not always make sense. For example, if your queue includes 50 messages but they only span 2 `MessageGroupId`s, then you will only be able to process 2 messages at a time and 48 worker machines will sit idle.
+
+For this reason, it's important that you plan for the number of `MessageGroupId`s you will use and set the template option `maxSize` accordingly. If you are only going to use 2 `MessageGroupId`s, you should set `maxSize: 2`.
+
+### Dead-letter queues and FIFO
+
+[The AWS documentation for dead-letter queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html) suggests the following:
+
+> Don’t use a dead-letter queue with a FIFO queue if you don’t want to break the exact order of messages or operations.
+
+If exact ordering is absolutely necessary for your program, you may prefer that the whole system stall if one message fails, instead of dropping the failed message into a dead-letter queue and moving on to the next one.
+
+Currently there is no way to turn off Watchbot's dead-letter queue. If you need this feature, please submit an issue or pull request.
--- a/lib/template.js
+++ b/lib/template.js
@@ -103,6 +103,11 @@ const dashboard = require(path.resolve(__dirname, 'dashboard.js'));
  * a number or a reference, i.e. `{"Ref": "..."}`.
  * @param {boolean} [options.deadletterAlarm=true] - Use this parameter to disable
  * creating an alarm for the dead letter queue threshold going from 0-1 messages.
+ * @param {Boolean} [options.fifo=false] - Whether you want Watchbot's SQS queue
+ * to be first-in-first-out (FIFO). By default, Watchbot creates a standard SQS queue,
+ * in which the order of jobs is not guaranteed to match the order of messages.
+ * If your program requires more precise ordering and the limitations of a FIFO queue
+ * will be acceptable, set this option to `true`.
  */
 module.exports = (options = {}) => {
   ['service', 'serviceVersion', 'command', 'cluster'].forEach((required) => {
@@ -131,7 +136,8 @@ module.exports = (options = {}) => {
       failedPlacementAlarmPeriods: 1,
       deadletterThreshold: 10,
       deadletterAlarm: true,
-      dashboard: true
+      dashboard: true,
+      fifo: false
     },
     options
   );
@@ -140,7 +146,7 @@ module.exports = (options = {}) => {
 
   const ref = {
     logGroup: cf.ref(prefixed('LogGroup')),
-    topic: cf.ref(prefixed('Topic')),
+    topic: options.fifo ? undefined : cf.ref(prefixed('Topic')),
     queueUrl: cf.ref(prefixed('Queue')),
     queueArn: cf.getAtt(prefixed('Queue'), 'Arn'),
     queueName: cf.getAtt(prefixed('Queue'), 'QueueName'),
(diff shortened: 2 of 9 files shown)


lib/template.js
```diff
@@ -461,7 +477,7 @@ module.exports = (options = {}) => {
     Type: 'AWS::ECS::Service',
     Properties: {
       Cluster: options.cluster,
-      DesiredCount: options.minSize,
+      DesiredCount: !options.fifo ? options.minSize : 1,
```
rclark: 👉 pinning the number of containers to `1`

lib/template.js
```diff
@@ -475,7 +491,7 @@ module.exports = (options = {}) => {
       options.placementStrategies;
 
 
-  Resources[prefixed('ScalingRole')] = {
+  if (!options.fifo) Resources[prefixed('ScalingRole')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -511,7 +527,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('ScalingTarget')] = {
+  if (!options.fifo) Resources[prefixed('ScalingTarget')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -528,7 +544,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('ScaleUp')] = {
+  if (!options.fifo) Resources[prefixed('ScaleUp')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -548,7 +564,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('ScaleUpTrigger')] = {
+  if (!options.fifo) Resources[prefixed('ScaleUpTrigger')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -567,7 +583,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('ScaleDown')] = {
+  if (!options.fifo) Resources[prefixed('ScaleDown')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -587,7 +603,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('ScaleDownTrigger')] = {
+  if (!options.fifo) Resources[prefixed('ScaleDownTrigger')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -734,7 +750,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('LambdaScalingRole')] = {
+  if (!options.fifo) Resources[prefixed('LambdaScalingRole')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -765,9 +781,9 @@ module.exports = (options = {}) => {
         }
       }]
     }
-  },
+  };
 
-  Resources[prefixed('ScalingLambda')] = {
+  if (!options.fifo) Resources[prefixed('ScalingLambda')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -785,7 +801,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('CustomScalingResource')] = {
+  if (!options.fifo) Resources[prefixed('CustomScalingResource')] = {
```
rclark: 👉 disabling autoscaling

lib/template.js
```diff
@@ -236,7 +244,13 @@ module.exports = (options = {}) => {
...
   };
 
-  Resources[prefixed('Topic')] = {
+  if (options.fifo) {
+    Resources[prefixed('Queue')].Properties.FifoQueue = true;
+    Resources[prefixed('Queue')].Properties.ContentBasedDeduplication = true;
+    Resources[prefixed('Queue')].Properties.QueueName = cf.join([cf.stackName, '-', prefixed('Queue'), '.fifo']);
+  }
+
+  if (!options.fifo) Resources[prefixed('Topic')] = {
```
rclark: 👉 remove the SNS topic

lib/template.js
```diff
@@ -248,7 +262,7 @@ module.exports = (options = {}) => {
     }
   };
 
-  Resources[prefixed('QueuePolicy')] = {
+  if (!options.fifo) Resources[prefixed('QueuePolicy')] = {
```
rclark: 👉 remove the SNS topic

davidtheclark:
@rclark I added some documentation to this branch. I think that the functionality is tested by the combination of template validation and snapshots. So this is ready for a review.

This branch seems to be working well in our prototype. But I'd be happy to sit on it until the functionality is more proven, if you're hesitant to merge into master.

docs/using-a-fifo-queue.md
```diff
@@ -0,0 +1,17 @@
+## Using a first-in-first-out (FIFO) queue
+
+By default, Watchbot creates a standard SQS queue, in which the order of jobs is not guaranteed to match the order of messages. If your program requires more precise ordering, you can use [a first-in-first-out (FIFO) queue](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html), instead.
+
+To use a FIFO queue, set the template option `fifo: true`.
+
+A FIFO queue behaves differently and has some limitations, so you should read the AWS documentation and the details below if you are considering the switch.
```
rclark: Could we link out here to what AWS documentation is helpful to read?

docs/using-a-fifo-queue.md
```diff
@@ -0,0 +1,17 @@
...
+
+### Triggering FIFO workers
+
+With a standard queue, you trigger a Watchbot worker by sending an SNS message to the watcher's topic. With a FIFO queue, you will send a message directly to the watcher's FIFO queue, instead. Make sure you have read about the AWS documentation about the expected for these messages (for example, you need to include a `MessageGroupId`).
```
rclark: > Make sure you have read **about** the AWS documentation about the expected for these messages

This sentence is a little clumsy. Additional `about` and `expected` vs `expectations` (I think?)

davidtheclark:
@rclark [the AWS docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html) for dead-letter queues suggest *not* using one for most FIFO queues: 

> Don’t use a dead-letter queue with a FIFO queue if you don’t want to break the exact order of messages or operations.

Do you think we should worry about this and add the ability to turn off the dead letter queue? Or should we just recommend that you should be cognizant and maybe set the `deadletterThreshold` to 1000?

rclark:
Perhaps we should get more clarity here, but generally I don't believe this suggestion. If you have a dead letter threshold of 10, then the FIFO queue should try that message 10 times and if it can't process, drop it in the dead letter queue. Now the FIFO queue can move on to the next message. Strictly speaking you are now out of order. This could be important for some applications where maybe the next message literally cannot proceed without the first one succeeding. But for many use cases, that may not be true -- the series of messages may not depend on each other but need ordering for another purpose.

So maybe there's a `strictFifo` option that doesn't have a dead letter queue? Maybe `fifo: 'strict'` or `fifo: 'lenient'` or something?

docs/using-a-fifo-queue.md
davidtheclark: Haha.

rclark:
> What do you think of ignoring this situation for now?

Sure: build it when the need arises.

rclark:
Published as v4.13.0

PR #280, opened 2018-08-21
configurable CPU alarm

brendanmcfarland:
Opinionated changes here, so other opinions very welcome

Changes:
 - **CPU Utilization Alarm can be configured**: Threshold and Periods can be set in the template
 - **Adjusted CPU Utilization Alarm period from 1 minute to 5 minutes**: Alarming on one minute CPU utilization spikes isn't a good signal of problematic behavior when the ECS environment is using `--cpu-shares`
 - **Adjusted Default CPU Alarm:** CPU alarm default now requires 200% average cpu utilization over four consecutive 5 minute periods to trigger an alarm
 - **Renamed alarm parameters**: Tried to remove ambiguity over which parameter belongs to which alarm; now that more alarms can be configured.

cc: @nickcordella


nickcordella:
Looks like it would be very helpful to some of my projects - thanks so much @brendanmcfarland !

lib/template.js
```diff
@@ -74,12 +83,12 @@ const dashboard = require(path.resolve(__dirname, 'dashboard.js'));
  * CloudWatch alarm that will fire if there have been more than this number
  * of failed worker invocations in a 60 second period. This parameter can be provided as
  * either a number or a reference, i.e. `{"Ref": "..."}`.
- * @param {Number|ref} [options.alarmThreshold=40] - Watchbot creates a
+ * @param {Number|ref} [options.messageAlarmThreshold=40] - Watchbot creates a
```
jakepruitt: This change will mean this is a breaking change 😕. Could we have `alarmThreshold` be an alias for `messageAlarmThreshold` for the time being to maintain backwards compatibility?

rclark: Related: do we really need the cpu alarms to be configurable? Or are the changes to the default sufficient?

brendanmcfarland: We could... remove the CPU utilization alarm completely and add a floor on CPU reservations. We can't accurately predict if any default CPU alarm will provide a signal for problematic usage; but we do know that 0 cpu reservations have caused bad behavior.

rclark: :+1: I am pro-simplification.

brendanmcfarland:
Closing in favor of #282

PR #281, opened 2018-08-21
[WIP] FIFO queue experimental changes

davidtheclark:
Merges into https://github.com/mapbox/ecs-watchbot/pull/279. I created a new branch so I could experiment with this for a little while without messing up the current `fifo` branch. 

cc @mapbox/frontend-platform @rclark

### DIFF
--- a/lib/message.js
+++ b/lib/message.js
@@ -29,14 +29,25 @@ class Message {
 
     if (!options.queueUrl) throw new Error('Missing options: queueUrl');
 
-    const snsMessage = JSON.parse(sqsMessage.Body);
+    // Only FIFO queue messages will have this property.
+    const messageGroupId = sqsMessage.MessageGroupId;
+    let envMessage;
+    let envSubject;
+    if (messageGroupId) {
+      envMessage = sqsMessage.Body;
+      envSubject = messageGroupId;
+    } else {
+      const snsMessage = JSON.parse(sqsMessage.Body);
+      envMessage = snsMessage.Message;
+      envSubject = snsMessage.Subject;
+    }
 
     this.id = sqsMessage.MessageId;
     this.handle = sqsMessage.ReceiptHandle;
 
     this.env = {
       MessageId: sqsMessage.MessageId,
-      Message: snsMessage.Message,
+      Message: envMessage,
       SentTimestamp: new Date(
         Number(sqsMessage.Attributes.SentTimestamp)
       ).toISOString(),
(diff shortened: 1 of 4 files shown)


lib/message.js
```diff
@@ -46,7 +46,10 @@ class Message {
       ApproximateReceiveCount: sqsMessage.Attributes.ApproximateReceiveCount.toString()
     };
 
-    if (snsMessage.Subject) this.env.Subject = snsMessage.Subject;
+    // Only FIFO queue messages will have this property.
+    const messageGroupId = sqsMessage.MessageGroupId;
+    const envSubject = messageGroupId || snsMessage.Subject;
+    if (envSubject) this.env.Subject = envSubject;
```
rclark: There's no guarantee that the SQS Message body will be `JSON.parse`able on L32 above, or that it will have a `.Message` property on L39.

davidtheclark: Ok. So I'm thinking now that the parsed body of the `sqsMessage` *is the `snsMessage`* because we make that happen, so when the FIFO SQS message is sent directly (instead of an indirect SQS message via SNS) we won't have that body. I'm having trouble seeing how the SNS & SQS hookup works, but I'll try running with that.

davidtheclark: Thanks. Ah, so that subscription must happen through the CloudFormation template that gets generated. I think I'm seeing now where that probably happens: https://github.com/mapbox/ecs-watchbot/blob/3975071b81a6381b4a93f74faa899a3b55046eee/lib/template.js#L239-L249

davidtheclark:
This seems to be working in a sense. However, I'm not currently seeing the message group behavior I'd expect. I set my worker's task to take 5 seconds, then fired off a series of messages in groups a1, a2, b1, a3, c1, b2. I expected these to end up processing in an order like a1, b1, c1, a2, b2, a3; but instead they all waited for each other, as though they were in the same group, and processed in the order that they were triggered. @rclark do you know of anything still left in this branch that would cause the queue to only allow one at a time, regardless of group? It's also possible that my quick test isn't quite right.

rclark:
Do you know how many containers you had running at the time of your test? If there was only 1, then this would make sense, since 1 container can only process 1 message at a time.

It can take watchbot up to 5 minutes to react to changes in the queue size and scale up the number of containers it is running. For your next test, you should either (a) make your jobs take considerably longer (like 5 minutes each), or (b) set the `MinSize` on your stack `>=` the number of `MessageGroupIds` you intend to queue up.

davidtheclark:
@rclark That was it, thanks. Added a minSize of 10 and then I saw the behavior that I expected.

davidtheclark:
Merging this info https://github.com/mapbox/ecs-watchbot/pull/279 to continue work.

PR #295, opened 2018-10-30
Add support for artifacts

vsmart:


vsmart:
@rclark can you think of a reason why we need image builds here? conex is subscribed but I see no Dockerfile except in `test/`, and we don't need to build that test image on each commit.

rclark:
Until v4 we had a dockerfile and needed it built on commit. No longer.

PR #300, opened 2018-11-27
Upgrade @mapbox/watchbot-progress

vsmart:
cc @rclark @drboyer


vsmart:
will bump the version after merging @drboyer !

PR #303, opened 2019-02-16
Add region to alarm names

rclark:
refs https://github.com/mapbox/platform/issues/1704
cc @mapbox/platform


rclark:
Nevermind, we shouldn't need to add region names to these scaling alarms.

PR #314, opened 2019-07-30
Document why the magic number 14 appears in the retry handler.

danpat:
Adds a comment to describe why the magic number 14 is used in the retry handler.

Thanks to @jakepruitt for pointing out where this magic number was discussed.


jakepruitt:
@danpat you're my hero

Issue #304, opened 2019-03-07
Empty sns subject not allowed since >= 4.13.0


vsmart:
Since watchbot 4.13.0 it is possible to use watchbot with a fifo queue and the message parsing was changed to accomodate this. This broke allowing empty sns subjects, which was originally introduced in https://github.com/mapbox/ecs-watchbot/pull/240. 

From [lib/message.js](https://github.com/mapbox/ecs-watchbot/blob/master/lib/message.js#L32-L49): 

    // If the Watchbot instance uses a regular SQS queue, sqsMessage
    // will have come from an SNS topic, so the body will be a JSON object
    // with Message and Subject properties.
    //
    // If the SQS message is for a FIFO queue, it did not come from SNS
    // so will not have the same structure and might not be JSON
    // parseable. If it is, we'll parse it; if not, we'll just pass it on.

We try to see if the message is parseable here: 

    try {
      const parsedBody = JSON.parse(sqsMessage.Body);
      if (parsedBody.Subject)  {
        envMessage = parsedBody.Message;
        envSubject = parsedBody.Subject;
      }
    } catch (error) {
      // JSON-parsing failure means we can leave the body as it is.
    }

However, this will fail if the message does not contain a subject. The `subject` is not a required key, and so even a standard sns message that came through a non-fifo queue but for some reason does not contain a `Subject` key will not be parsed as expected.

I ran into this this week with two separate repositories while upgrading them to the latest watchbot version - in one case, I was sending an SNS message via the console and did not specify a Subject (again, not a required parameter), and in another case, I was sending the message using code like this: 

```
sns
    .publish({
      Message: "my message",
      TopicArn: "arn:aws:sns..." 
})
```

... without specifying a `Subject`. 

## Next actions

I could imagine two possible fixes, but I need some more context on whether those might break something else 🤔 

1. Explicitly check that `option.fifo === true` in the `try` statement - since if I understand correctly, this is only needed for fifo queues. 
2. remove the `if parsedBody.Subject` clause and always assign message to `parseBody.Message`  if `parsedBody.Message` exists and the json parsing was successful 

cc @mapbox/platform

PR #315, opened 2019-08-06
Note "Subject" is not optional for SNS messages

sgillies:
Since 4.13.0. Specifically because of https://github.com/mapbox/ecs-watchbot/pull/279/files#diff-cb88e96d819797caf0f16e3e34843499R43.

Related to https://github.com/mapbox/pxm/issues/1345, https://github.com/mapbox/ecs-watchbot/issues/304.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -27,7 +27,8 @@
 
 ### 4.13.0
 
-- Adds support for first-in-first-out (FIFO) SQS queues. https://github.com/mapbox/ecs-watchbot/pull/279
+- Adds support for first-in-first-out (FIFO) SQS queues. https://github.com/mapbox/ecs-watchbot/pull/279.
+- SNS message objects are required to have both a "Message" and a "Subject" property. Previously, "Subject" was optional.
 
 ### 4.12.0
 - Add `options.deadletterAlarm` (default=true) to disable the alarm resource for dead letter queue messages https://github.com/mapbox/ecs-watchbot/pull/288


changelog.md
```diff
@@ -27,7 +27,8 @@
 
 ### 4.13.0
 
-- Adds support for first-in-first-out (FIFO) SQS queues. https://github.com/mapbox/ecs-watchbot/pull/279
+- Adds support for first-in-first-out (FIFO) SQS queues. https://github.com/mapbox/ecs-watchbot/pull/279.
+- SNS message objects are required to have both a "Message" and a "Subject" property. Previously, "Subject" was optional.
```
vsmart: Could you link to the issue ticket here? This wording sounds like it was an intentional change, but in fact it was a regression.

drboyer: Just so I understand, subjects used to be required, a regression was introduced at some point and it was fixed at version 4.13.0?

vsmart: No. Subject used to be optional. Now they are required, i.e. the regression is that we accidentally introduced this requirement. More details at #304.

sgillies: @vsmart the Travis errors were transient, I re-ran the builds and all is green.

Point taken about this being more of a regression. I don't usually edit my own change logs to retroactively note regressions introduced, only their fixes. Feel free to close this, though I would love to see *some* action. 3 different Mapbox projects stumbled over this.

PR #316, opened 2019-08-20
Dashboard: Add queue oldest-message wait time, worker duration to CW dashboard

mookerji:
This surfaces two metrics aggregates to the Cloudwatch dashboard:

- wait time of the oldest message in the queue
- worker duration

Each of these can be used as a service level indicator for job queueing backlog
in an async API.

/cc @rclark @drboyer
/fyi @danpat


mookerji:
Looks like I need to follow up with fixing broken tests! Will do so tomorrow.

PR #317, opened 2019-08-21
FIX: Missing properties/type key in CW dashboard change.

mookerji:
Fixes:
- https://github.com/mapbox/ecs-watchbot/pull/316

/cc @jakepruitt


mookerji:
Looks like we missed this @jakepruitt; I cut a release and was forced to rollback a deploy after the dashboard resource failed schema checks. I checked a rendered template manually on CW and it seems good. Follow up is to release a new patch version after this PR is merged.

mookerji:
Verified that the template is fixed with a deployed app. Merging and cutting a patch release.

PR #319, opened 2019-09-05
Expose SQS visibility timeout to Cloudformation template.

mookerji:
In the event that a task is externally terminated, a fixed 180 seconds
must pass before the work item is visible again, which may be too long
for some applications. This permits watchbot apps to define their own
visibility timeout for work deadlines.

/cc @rclark @jakepruitt
/fyi @danpat

### DIFF
--- a/docs/building-a-template.md
+++ b/docs/building-a-template.md
@@ -89,7 +89,9 @@ When creating your watchbot stacks with the `watchbot.template()` method, you no
 **errorThreshold** | Watchbot creates a CloudWatch alarm that will fire if there have been more than this number of failed worker invocations in a 60 second period. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. | Number/Ref | No | 10
 **deadletterThreshold** | Use this parameter to control the number of times a message is delivered to the source queue before being moved to the dead-letter queue. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. | Number/Ref | No | 10
 **dashboard** | Watchbot creates a Cloudwatch Dashboard called `<cloudformation-stack>-<region>`. If running in cn-north-1, this may need to be disabled | Boolean | No | `true`
-**fifo** | Whether you want Watchbot's SQS queue to be first-in-first-out (FIFO). By default, Watchbot creates a standard SQS queue, in which the order of jobs is not guaranteed to match the order of messages. If your program requires more precise ordering and the limitations of a FIFO queue will be acceptable, set this option to `true`. Learn more in ["Using a FIFO queue"](./using-a-fifo-queue.md) | Boolean | No | `false`
+**fifo** | Whether you want Watchbot's SQS queue to be first-in-first-out (FIFO). By default, Watchbot creates a standard SQS queue, in which the order of jobs is not guaranteed to match the order of messages. If your program requires more precise ordering and the limitations of a FIFO queue will be acceptable, set this option to `true`. Learn more in ["Using a FIFO queue"](./using-a-fifo-queue.md). | Boolean | No | `false`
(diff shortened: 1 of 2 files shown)


mookerji:
yo, @rclark @jakepruitt I have a question about how visibility timeout interacts with out backoff strategy here. Specifically this block: 

https://github.com/mapbox/ecs-watchbot/blob/master/lib/message.js#L79..L86

- When doing backoff, should we multiply the backoff interval by the original visibility timeout? i.e., the new visibility timeout on retries should instead be: `Max(MAX_VISIBILITY_TIMEOUT, defaultVisibilityTimeout*Math.pow(2, num_receives))`. Otherwise, starting with the current 180 second default visibility timeout, the new visibility timeout after a retry (with `num_receives=1`) will be **less** than the visibility timeout from the first watchbot execution, which could trigger a duplicate task.
- There a comment in that block about the visibility timeout being `<2^15` because of AWS rate limits, but that appears not to be the case? `2^15 sec / 3600 sec / hour ~ 9 hours`. But the actual max is [12 hours](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-limits.html). Regardless, we should check that `defaultVisibilityTimeout*Math.pow(2, num_receives)) < MAX_VISIBILITY_TIMEOUT`?

If those points make sense, let me know and I can add them to this PR (or make another one).

rclark:
Or maybe derive the heartbeat intervals from the `maxJobDuration` option?

mookerji:
@rclark oh yeah, I agree that `visibilityTimeout` is actually an implementation detail, and what we really should expose here is the configuration around heartbeats. I'll have a think about it.

Did you have any thoughts around the computing of the backoff interval and whether that should include visibility?

rclark:
The interval doesn't need to be multiplied because [the retry() function is only ever called after a worker process errored](https://github.com/mapbox/ecs-watchbot/blob/3c5fd1730178560c745fd792dc9fcf5862876d3f/lib/worker.js#L106). Since the first one failed there's no risk of overlapping processing.

As for the 2^14, it never quite got elucidated [here](https://github.com/mapbox/ecs-watchbot/pull/184/files#r167034116), as the code comment suggests. However, I think this is actually a heartbeat-retry mixup and a lurking bug.

[In this cryptic AWS documentation](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html?shortFooter=true#changing-message-visibility-timeout)

> The new timeout period takes effect from the time you call the ChangeMessageVisibility action.

So say you receive a message with 3 min timeout. 2 min later, you call the API to change the message visibility by 2 min [like we do with the heartbeat](https://github.com/mapbox/ecs-watchbot/blob/3c5fd1730178560c745fd792dc9fcf5862876d3f/lib/message.js#L96-L106). The message is now invisible for **5 minutes**. Since there is accumulation, you can't get to 15 **heartbeats** (2^14 + 2^15 seconds > 12 hours). But we don't check this at all when we're doing our heartbeats.

We do check it on message retry(), which is only after an explicit worker error. In this case we **should** be able to get to 2^15. That said, should we? The [default receives before drop to dead letter is 10](https://github.com/mapbox/ecs-watchbot/blob/3c5fd1730178560c745fd792dc9fcf5862876d3f/lib/template.js#L100) and I might advocate we cap it there. If it failed 10 times, something else is probably wrong.

As for bugs: when we call retry() we actually may have already heartbeated the timeout to its maximum value, or even [errored in an unhandled fashion on several setIntervals](https://github.com/mapbox/ecs-watchbot/blob/3c5fd1730178560c745fd792dc9fcf5862876d3f/lib/worker.js#L90). In this case [the call to ChangeMessageVisibility](https://github.com/mapbox/ecs-watchbot/blob/3c5fd1730178560c745fd792dc9fcf5862876d3f/lib/message.js#L90) would fail even on receive count = 1.

jqtrde:
Stale - going to close this one out.

Issue #324, opened 2019-12-10
Node 10 support


hcourt:
Node 8 is EOL on 12/31/2019.
The stacks created by this repository are running on Node 8 and should be upgraded to Node 10.

ScalingLambda:
https://github.com/mapbox/ecs-watchbot/blob/f07922224b243539deb1cdb62463fe42c74e17eb/lib/template.js#L833

TotalMessagesLambda:
https://github.com/mapbox/ecs-watchbot/blob/f07922224b243539deb1cdb62463fe42c74e17eb/lib/template.js#L881

mapsam:
@hcourt this is a good catch! I'm curious, with the code freeze starting on 12/16 and the node8 lambdas EOL coming up, are teams using ecs-watchbot expected to get their stacks updated this week?

If not - is it acceptable risk to wait to update stacks until after January 6th? (giving us one month to make updates to functions before the "no update" date on February 3rd)?

hcourt:
@mapsam  Good question!

I think for teams that the code freeze applies to, they should be able to get approval if they need it.  If I'm able to merge this today, then this week would be obviously a much better time than next week to make the change.

January 6th is the date after which AWS will disable creating Lambdas with Node 8 runtimes.  The risk here is obviously if your Lambda is accidentally deleted.  If your team does not wish to go through the CM process during the freeze, then I would advise updating your staging stacks to Node10 in preparation for the update after the freeze.  This way you can be sure to have the change ready and on master.

PR #325, opened 2019-12-10
Node 10 support

hcourt:
Fixes https://github.com/mapbox/ecs-watchbot/issues/324


.artifacts.yml
```diff
@@ -4,4 +4,4 @@ bundle:
     - filter: on-every-commit
       config:
         platform: nodejs
-        platform_version: '8.10'
+        platform_version: '10'
```
vsmart: You could also opt to move to the default bundle, which is Node 10. That artifacts.yml would be:
```
version: v1
defaults: 
- bundle
```

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "engines": {
-    "node": ">=8"
```
vsmart: Could you also update some of the Node dependencies as part of the release? E.g. mapbox/cloudfriend is 3 major versions behind. `npm outdated` will give a rundown of the major outdated ones.

hcourt: Doing so here: https://github.com/mapbox/ecs-watchbot/pull/326

hcourt:
Tested with a staging stack.  Logs indicated that messages could be processed successfully.

hcourt:
One (custom) resource was failing to finish creation, and I believe this is due to https://github.com/aws/aws-sdk-js/issues/2955

Trying out a fix for this.

hcourt:
> One (custom) resource was failing to finish creation, and I believe this is due to [aws/aws-sdk-js#2955](https://github.com/aws/aws-sdk-js/issues/2955)
> 
> Trying out a fix for this.

This worked.

hcourt:
Think this is finished.  Before releasing, I will also merge #326.

lib/template.js
```diff
@@ -823,14 +823,14 @@ module.exports = (options = {}) => {
       Role: cf.getAtt(prefixed('LambdaScalingRole'), 'Arn'),
       Code: {
         ZipFile: cf.sub(`
-          const response = require('cfn-response');
+          const response = require('./cfn-response');
```
vsmart: Wow - nice find, though that is super weird. I did see that in the support ticket you linked (https://github.com/aws/aws-sdk-js/issues/2955) they also recommend trying to update the AWS-SDK node dependency to the latest version. Would you consider trying that to see if it makes this change not needed?
We could also ship this as-is and open a support case to track - I'll leave that up to you.

hcourt: Hm, going to try it out, shouldn't take that long.

hcourt: I think this is a matter of where the Lambda or CloudFormation code packages `cfn-response`, not anything to do with the aws sdk.  The support ticket I linked also mentions this.

I did a lot of testing but I'm ready to let this go for now.  I think even using a different package version would be more onerous than this pretty simple workaround.  I will open a support case to track it.

PR #328, opened 2020-01-02
Functional Alpine binaries

### LINKED ISSUE #327: v4.20.0 binary on alpine causes segmentation fault
Seeing that [ecs-telephone](https://github.com/mapbox/ecs-telephone/tree/420) with ecs-watchbot version 4.20.0 is failing with exit code `137` repeatedly:

Screen Shot 2019-12-19 at 5 27 43 PM

We should dig into whether the issue is in the watchbot binary or in one of the dependencies. If we can't find the answer by the end of the year, we might want to roll back the 4.20 version.

rclark:
This PR adjusts the process that builds binaries for new ecs-watchbot releases.

- Adds a 2nd CodeBuild project to the CodePipeline. This project runs on an Alpine container which allows it to produce a functioning Alpine binary. The Alpine build only makes the one binary, the other three builds (linux, windows, macos) are still performed as they were before on amazon linux. This change fixes #327.

- Removes the CodePipeline dependency on https://github.com/rclark/code-pipeline-helper, opting instead to use [dynamic references to secrets stored in secrets manager](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html#dynamic-references-secretsmanager) to authorize AWS access to this GitHub repository.

cc @shwetha-mc @hcourt

### DIFF
--- a/bin/watchbot-binary-generator.js
+++ b/bin/watchbot-binary-generator.js
@@ -58,7 +58,7 @@ const uploadBundle = async () => {
   const sha = process.env.CODEBUILD_RESOLVED_SOURCE_VERSION;
 
   const tag = await getTagForSha(sha);
-  if (process.argv[2] === 'alpine' && tag) {
+  if (tag) {
     const uploads = targets.map((target) => {
       console.log(`Uploading the package to s3://${Bucket}/${target.prefix}/${tag}/watchbot`);
       return s3.putObject({
--- a/bin/watchbot-binary-generator.js
+++ b/bin/watchbot-binary-generator.js
@@ -58,7 +58,7 @@ const uploadBundle = async () => {
   const sha = process.env.CODEBUILD_RESOLVED_SOURCE_VERSION;
 
   const tag = await getTagForSha(sha);
-  if (tag) {
+  if (process.argv[2] === 'alpine' && tag) {
     const uploads = targets.map((target) => {
       console.log(`Uploading the package to s3://${Bucket}/${target.prefix}/${tag}/watchbot`);
       return s3.putObject({
--- a/bin/watchbot-binary-generator.js
+++ b/bin/watchbot-binary-generator.js
@@ -36,7 +36,7 @@ wbg.getTagForSha = getTagForSha;
 /*
  * uploadBundle - uploads watchbot binaries to S3
  */
-const uploadBundle = async (alpineFlag) => {
+const uploadBundle = async (buildTarget) => {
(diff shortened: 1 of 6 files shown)


bin/watchbot-binary-generator.js
```diff
@@ -36,17 +36,22 @@ wbg.getTagForSha = getTagForSha;
 /*
  * uploadBundle - uploads watchbot binaries to S3
  */
-const uploadBundle = async () => {
+const uploadBundle = async (alpineFlag) => {
```
rclark: Let's either (a) change this to a boolean value or (b) change the variable name to `target`. Maybe someday we'd want to 

```
node bin/watchbot-generate-binaries win
```

... or something.

bin/watchbot-binary-generator.js
```diff
@@ -72,11 +77,20 @@ const uploadBundle = async () => {
...
+        process.exit(1);
+      });
+  }
+  else {
+    uploadBundle()
+      .catch((err) => {
+        console.log(err);
+        process.exit(1);
+      });
+  }
```
hcourt: Because of the way you've written `uploadBundle`, I think you can safely just pass `process.argv[2]` without checking the value.  `uploadBundle` will ignore it if it isn't `alpine`.

cloudformation/ecs-watchbot-generate-binaries.template.js
```diff
@@ -134,20 +177,60 @@ const Resources = {
...
+                Provider: 'GitHub'
+              },
+              OutputArtifacts: [
+                { Name: 'Source' }
+              ],
+              Configuration: {
+                Owner: 'mapbox',
+                Repo: 'ecs-watchbot',
+                PollForSourceChanges: 'false',
+                Branch: 'alpine-binaries',
```
hcourt: Just a reminder to update this when you're done testing.

bin/watchbot-binary-generator.js
shwetha-mc: Good idea! Changing!

bin/watchbot-binary-generator.js
shwetha-mc: Thanks, @hcourt!

vsmart:
@shwetha-mc Is this ready for release as 4.20.1?

shwetha-mc:
Great question, @vsmart! Publishing now.

PR #342, opened 2020-07-08
Apply prefix to ClusterArn

### LINKED ISSUE #341: Cluster parameter does not respect prefixes
I'm using the `prefix` option to deploy multiple watchbot services using a [single CFN stack](https://github.com/mapbox/tiledown/blob/split-reduce/cloudformation/tiledown.template.js). This works fine for most parameters, but trying to set different values for the `cluster` results in an error:

```
Error: Outputs name used more than once: ClusterArn
```

The ClusterArn [appears](https://github.com/mapbox/ecs-watchbot/blob/0c19139f6e7a2ca76c07606af87ad48017e8d473/lib/template.js#L953) to be treated differently than the other options in that it isn't prefixed, so defining multiple watchbot services in a single template fails unless the cluster is always the same. Is having multiple clusters within a single stack something that watchbot can support?

michaelskahn:
Use the specified `prefix` parameter when creating the ClusterArn outputs.

This is a breaking change for any consumer using the current ClusterArn output key. The previous behavior always had a `ClusterArn` output key, and the template would fail to build with multiple distinct clusters defined. The new behavior is:
* If no prefix is specified, the default output key is `WatchbotClusterArn`
* If prefixes are specified, an output key is created for each prefix in the pattern `{prefix}ClusterArn`

Also, add node-modules to .gitignore.

Fixes #341

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .nyc_output/
 coverage/
 vendor/
+node_modules
--- a/lib/template.js
+++ b/lib/template.js
@@ -949,11 +949,11 @@ module.exports = (options = {}) => {
     }
   };
 
-  const outputs = {
-    ClusterArn: {
-      Description: 'Service cluster ARN',
-      Value: options.cluster
-    }
+  const outputs = {};
+
+  outputs[prefixed('ClusterArn')] = {
+    Description: 'Service cluster ARN',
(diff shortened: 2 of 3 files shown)


michaelskahn:
I'll note that we have a workaround (renaming values before merging), so this isn't an urgent fix.

PR #344, opened 2020-10-09
Add Fargate and Fargate Spot compatibility

davidtheclark:
This PR proposes a new `fargate` options object that allows users to run Watchbot tasks on Fargate or Fargate Spot capacity.

An interface challenge here is that Fargate is incompatible with several of the other options, so I added some documentation and validation to make that clear. Several of the incompatible options were not already documented, so I added those docs.

I have not yet tried launching a test stack with this, but decided to put up this draft PR to see if anybody has feedback about the interface changes. If the interface looks good to people, then the following steps are still required:

- Unit tests for validation of incompatible options.
- Validate that it works in real life.

cc @mapbox/api-infrastructure


docs/building-a-template.md
```diff
@@ -70,16 +70,21 @@ When creating your watchbot stacks with the `watchbot.template()` method, you no
...
+**reservation.memory** | The number of MB of RAM to reserve as a hard limit. If your worker container tries to utilize more than this much RAM, it will be shut down. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. If you use `fargate`, this option must be set, and must be an allowed value for Fargate tasks. | Number/Ref | No | none
+**reservation.softMemory** | The number of MB of RAM to reserve as a soft limit. Your worker container will be able to utilize more than this much RAM if it happens to be available on the host. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. If you use `fargate`, this option cannot be set. | Number/Ref | No | none
+**reservation.cpu** | The number of CPU units to reserve for your worker container. This will only impact the placement of your container on an EC2 with sufficient CPU capacity, but will not limit your container's utilization. The minimum value is 128. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. If you use `fargate`, this option must be set, and must be an allowed value for Fargate tasks. | Number/Ref | Yes | -
```
rclark, line 85: What do you think about wiring default values for you if you chose Fargate? It'd be appropriate to provide the minimum valid cpu/memory as a default, I think.

docs/building-a-template.md
```diff
@@ -91,6 +96,14 @@ When creating your watchbot stacks with the `watchbot.template()` method, you no
...
+**placementConstraints** | ECS service [placement constraints](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-placementconstraint.html). If you use `fargate`, this option cannot be set. | Object[]/Ref | No | false
+**placementStrategies** | ECS service [placement strategies](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-service-placementstrategy.html). If you use `fargate`, this option cannot be set. | Object[]/Ref | No | false
+
+### fargate mode explained
+
+If you use the `fargate` option, Watchbot will launch tasks on Fargate capacity, rather than EC2 capacity.
+
+The choice to use `fargate` affects several other options above. Not all ECS task and service options are available on Fargate, and some must be specified. For example, the options `reservations.softMemory`, `privileged`, `writableFileSystem`, `placementConstraints`, and `placementStrategies` cannot be used alongside `fargate`. `reservations.memory` and `reservations.cpu` must be set, and they must be [values that are allows for Fargate](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html).
```
rclark, line 106:
```suggestion
The choice to use `fargate` affects several other options above. Not all ECS task and service options are available on Fargate, and some must be specified. For example, the options `reservations.softMemory`, `privileged`, `writableFileSystem`, `placementConstraints`, and `placementStrategies` cannot be used alongside `fargate`. `reservations.memory` and `reservations.cpu` must be set, and they must be [values that are allowed for Fargate](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html).
```

docs/building-a-template.md
davidtheclark: Yeah, that sounds good.

davidtheclark:
I tried this out on a real stack and it worked (after b8e6b25).

This is ready for a review from repo-owners @mapbox/data-tools, please. If it looks good, I'll cut a beta release and try it one more time before release.

davidtheclark:
Hm, I think I that I need to do a better job with some of these boolean options. I'll comment back when I've fixed that.

davidtheclark:
OK, I have a new API proposal: 

- `capacity` can be `EC2`, `FARGATE`, or `FARGATE_SPOT`, defaulting to `EC2`.
- `fargateSecurityGroups` and `fargateSubnets`.

I've changed a bunch of things from the previous so we use native CFN `Conditions` and `Fn::If` statements instead of JS conditions. The problem I've trying to address is that JS conditions are not able to deal effectively with the full range of possible values *if we want to allow options to intrinsic functions*. I think there's more room for this kind of improvement in the library, but I limited the changes to properties that are affected by `capacity`.

One effect of this is that the template hardly changes at all based on your `capacity` value. Just the `Conditions` change: the rest stays the way it was, with the same `Fn::If` functions that use those `Conditions`. Another is that *all* the existing snapshots change. They should still *resolve* to the same values they had before; but the template JSON itself now includes conditions and `Fn::If` functions it didn't contain before.

I've tried this out, launching with each of the 3 types of `capacity`, and it seems to be working well. 

I updated readme documentation, and before updating the JSDoc docs I wanted to ask: What would you think about me deleting those JSDoc comments that duplicate the Markdown? If we aren't going to generate the Markdown automatically from JSDoc comments, it seems like it could be better to have just one place to write down this information, to reduce duplication problems.

So this is ready for review @mapbox/data-tools (and @rclark if you'd like to take another look). Let me know if there's any testing you'd particularly like me to try.

lib/template.js
```diff
@@ -144,6 +177,12 @@ module.exports = (options = {}) => {
 
   const prefixed = (name) => `${options.prefix}${name}`;
 
+  options.reservation.memory = options.reservation.memory || cf.if(prefixed('CapacityIsNotEC2'), 512, cf.noValue);
+  options.reservation.cpu = options.reservation.cpu
+    ? Math.max((options.reservation.cpu || 0), 128)
+    : cf.if(prefixed('CapacityIsNotEC2'), 256, Math.max((options.reservation.cpu || 0), 128));
```
drboyer, line 183: This condition might need a little additional work, especially if we want to support passing this value dynamically (via a `{ Ref: 'x' }` or similar). I'm not sure if anyone is using references with these properties currently, but it seems conceivable that they might want to.

It looks like this is saying "if the cpu reservation is set, set the reservation to the greater of the cpu reservation or 128. If it's not set, set the value to 256 if the capacity type is not EC2, else set the reservation to the greater of the cpu reservation or 128". There are two problems here:

1. The last part of that is kinda nonsensical if we already know `options.reservation.cpu` is not set. Can we just set it to 128 or some other value?
2. This will break if options.reservation.cpu is set to a non-numeric value:

```js
> prop = { Ref: 'value' }
{ Ref: 'value' }
> Math.max((prop || 0), 128)
NaN
> prop = 164
164
> Math.max((prop || 0), 128)
164
```

The memory properties don't suffer from this problem. I also don't entirely understand the historic minimum of 128 units. That doesn't [appear to be a ECS-imposed restriction](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_environment). Can we just copy what we do with `options.reservation.memory`, set the default to 256, and call it a day?

lib/template.js
```diff
@@ -390,6 +438,46 @@ module.exports = (options = {}) => {
       }
     });
 
+  Resources[prefixed('ExecutionRole')] = {
```
drboyer, line 441: this is provided by the IAM role on the EC2 host for the `EC2` launch type?

lib/template.js
```diff
@@ -473,26 +568,29 @@ module.exports = (options = {}) => {
...
+          }
+        },
+        cf.noValue
+      ),
+      CapacityProviderStrategy: cf.if(
+        prefixed('CapacityIsFargateSpot'),
+        [{ CapacityProvider: 'FARGATE_SPOT', Weight: 1 }],
+        cf.noValue
+      ),
+      LaunchType: cf.if(prefixed('CapacityIsFargate'), 'FARGATE', cf.noValue)
```
drboyer, line 593: Asking this because I haven't worked with Fargate containers yet: So you must set `CapacityProviderStrategy` if using Fargate Spot, but you don't need to set `LaunchType` in that case? And if using regular Fargate, you don't need to set a `CapacityProviderStrategy`?

test/__snapshots__/template.spec.js.snap
```diff
@@ -1160,7 +1310,59 @@ Object {
             ],
           },
         ],
+        "Cpu": Object {
+          "Fn::If": Array [
+            "SoupCapacityIsNotEC2",
+            NaN,
```
drboyer, line 1316: you can see the problem I identified above here

docs/building-a-template.md
```diff
@@ -70,16 +70,20 @@ When creating your watchbot stacks with the `watchbot.template()` method, you no
 **serviceVersion** | The version of your image to deploy. This should reference a specific image in ECR. | String/Ref | Yes | -
 **family** | The name of the task definition family that watchbot will create revisions of. | String/Ref | No | -
 **command** | The shell command to be run by the subprocess worker. The working directory for the subprocess is determined in your Dockerfile by the `WORKDIR` missive. | String | Yes | -
+**capacity** | If the value is `'EC2'` (the default), Watchbot will run with the `EC2` launch type. If the value is `FARGATE`, Watchbot will run with the `FARGATE` launch type. And if it is `FARGATE_SPOT`, Watchbot will run on `FARGATE_SPOT` capacity provider. You can only use `FARGATE_SPOT` on ECS clusters configured to allow the `FARGATE_SPOT` capacity provider. | `'EC2'|'FARGATE'|'FARGATE_SPOT'`/Ref | No | `'EC2'`
```
drboyer, line 73: nit: I feel like "Watchbot will run with the _x_ launch type" isn't as clear as it could be. Consider something like `Watchbot will be deployed to ECS using the _x_ launch type`.

docs/building-a-template.md
```diff
@@ -70,16 +70,20 @@ When creating your watchbot stacks with the `watchbot.template()` method, you no
...
+**fargateSecurityGroups** | The IDs of security groups to associate with Fargate Watchbot containers. This option is ignored if `capacity` is `'EC2'`. If you do not specify a security group and use a Fargate capacity type, the default security group for the VPC is used. You can provide no more than 5. | String[]/Ref | No | []
+**fargateSubnets** | The IDs of subnets to associate with Fargate Watchbot containers. This option is ignored if `capacity` is `'EC2'`. You can provide no more than 16. | String[]/Ref | No | []
```
drboyer, line 75: Can we note somewhere on this page where one could get these values from?

docs/building-a-template.md
```diff
@@ -70,16 +70,20 @@ When creating your watchbot stacks with the `watchbot.template()` method, you no
...
-**privileged** | Give the container elevated privileges on the host container instance | Boolean | No | false
+**reservation.cpu** | The number of CPU units to reserve for your worker container. This will only impact the placement of your container on an EC2 with sufficient CPU capacity, but will not limit your container's utilization. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. The minimum value is 128 if you do not use `fargate`. If `capacity` is `'FARGATE'` or `'FARGATE_SPOT`', this must be an allowed value for Fargate tasks. | Number/Ref | Yes | None if `capacity` is `'EC2'`; `256` if it is `'FARGATE'` or `'FARGATE_SPOT`'
+**privileged** | Give the container elevated privileges on the host container instance. This value is ignored if `capacity` is not `'EC2'`. | Boolean/Ref | No | false
```
drboyer, line 85: For the 4 options where the value is ignored on Fargates, how would you feel about wording the 2nd sentence as "This value is ignored for capacity values other than `EC2`"? I feel like the "ignored if x is not y" is slightly confusing to me. Maybe not using the word "not" is slightly better?

drboyer (commented):
Couple questions and suggestions here. I think the biggest thing to fix is the Cpu reservation default behavior, as I explain below.

Overall, relying on CloudFormation conditions instead of just JS functionality sounds good to me - will likely make writing templates a bit easier.

drboyer:
> I updated readme documentation, and before updating the JSDoc docs I wanted to ask: What would you think about me deleting those JSDoc comments that duplicate the Markdown? If we aren't going to generate the Markdown automatically from JSDoc comments, it seems like it could be better to have just one place to write down this information, to reduce duplication problems.

That seems reasonable to me. I know the @mapbox/cloudfriend package uses the [`documentation`](https://www.npmjs.com/package/documentation) package to [autogenerate shortcut documentation](https://github.com/mapbox/cloudfriend/blob/196178c17b450e674754db95260ba96a4d8c74a9/package.json#L15). So it seems we might be able to go the other way as well. Up to you though, I don't have a strong preference.

lib/template.js
davidtheclark: > Can we just copy what we do with options.reservation.memory, set the default to 256, and call it a day?

Sounds great to me.

lib/template.js
davidtheclark: It grants the ECS container agent permission to make some AWS API calls. For Fargate (with or without Spot) it is required if you are pulling the image from ECR or publishing to CW logs.

lib/template.js
davidtheclark: It is confusing, yes. Capacity providers replace launch types: they are not compatible. There's no `launchType` way to use Fargate Spot: you have to use capacity providers. There is an established `launchType` way to use Fargate (non-Spot). You could also use a `FARGATE` capacity provider, but the capacity provider requires that you add some configuration to your cluster, so I thought I'd go with `launchType` for non-Spot Fargate.

docs/building-a-template.md
davidtheclark: I don't think so, because this is a public repo. We might need to add an internal guide about how to use this.

davidtheclark:
If you'd like, I'll cut the JSDoc comments for now and we'll stick with just the Markdown docs. If you are interested in autogenerating Markdown from the JSDoc I can share some tips about how to do that, but would rather not add that to this PR.

lib/template.js
davidtheclark: I think I'm going to replace this custom role with a reference to the AWS-provided task execution role: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_execution_IAM_role.html. The only different is it applies its CloudWatch logs permissions to a `*` instead of specific log groups.

davidtheclark:
Thanks for the careful review, @drboyer. I addressed the feedback in 8b3524f. If you have more feedback, let me know. Once things are suitable here, I'll try my manual test again to make sure it still works as expected.

lib/template.js
drboyer: sgtm, the less duplicative roles we create, the [better](https://github.com/mapbox/aws-permissions/issues/1)

lib/template.js
drboyer: Gotcha, you're certainly more knowledgeable than me about how to launch tasks onto the various Fargates, so I trust your judgement. Thanks for explaining the design decision in more detail here.

docs/building-a-template.md
drboyer: d'oh, right. Internal guide might be helpful to fill these in.

docs/building-a-template.md
```diff
@@ -70,16 +70,20 @@ When creating your watchbot stacks with the `watchbot.template()` method, you no
...
-**reservation.cpu** | The number of CPU units to reserve for your worker container. This will only impact the placement of your container on an EC2 with sufficient CPU capacity, but will not limit your container's utilization. The minimum value is 128. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. | Number/Ref | Yes | -
-**privileged** | Give the container elevated privileges on the host container instance | Boolean | No | false
+**reservation.cpu** | The number of CPU units to reserve for your worker container. This will only impact the placement of your container on an EC2 with sufficient CPU capacity, but will not limit your container's utilization. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. If `capacity` is `'FARGATE'` or `'FARGATE_SPOT`', this must be an allowed value for Fargate tasks. | Number/Ref | Yes | None if `capacity` is `'EC2'`; `256` if it is `'FARGATE'` or `'FARGATE_SPOT`'
```
drboyer, line 84:
```suggestion
**reservation.cpu** | The number of CPU units to reserve for your worker container. This will only impact the placement of your container on an EC2 with sufficient CPU capacity, but will not limit your container's utilization. This parameter can be provided as either a number or a reference, i.e. `{"Ref": "..."}`. If `capacity` is `'FARGATE'` or `'FARGATE_SPOT`', this must be an allowed value for Fargate tasks. | Number/Ref | Yes | `128` if `capacity` is `'EC2'`; `256` if `capacity` is `'FARGATE'` or `'FARGATE_SPOT`'
```

drboyer:
> If you'd like, I'll cut the JSDoc comments for now and we'll stick with just the Markdown docs. If you are interested in autogenerating Markdown from the JSDoc I can share some tips about how to do that, but would rather not add that to this PR.

Cool, seems totally reasonable to do that in a separate PR.

---

I looked over the PR one more time, and found only one other small docs correction. Other than that this seems pretty good to go. Feel free to do another test and report back!

davidtheclark:
I released a beta version and used that in another test. It worked great 👍

drboyer (approved):
I also reviewed the test case @davidtheclark sent me. Looks good to me!

@davidtheclark were you planning to release this as a new major version or minor version? I'd consider leaning towards major just because the new capacity options do represent a pretty significant change. The `reservations.cpu` change is more of a _breaking_ API change, also.

Also let me know if you're not comfortable running the new version release process yourself.

PR #345, opened 2020-10-19
Document more about switching capacity values

davidtheclark:
I think this is helpful information for anybody starting to use the `capacity` option. Might save them some trial and error. @drboyer for review, please.

cc @mapbox/api-infrastructure


drboyer (approved):
Thanks for the extra details @davidtheclark, this does sound helpful

PR #346, opened 2020-10-27
Add fargatePublicIp option

davidtheclark:
We want the ability to enable public IP addresses on `awsvpc` mode tasks.

@drboyer for review, please.


drboyer (approved):
We discussed the logic behind this change and the default value outside this PR and concluded this was reasonable. The change also looks technically correct.

PR #348, opened 2021-01-21
Note that Subject and Message are parsed

jseppi:
The fact that the Subject and Message are already parsed in the worker's `process.env` caught me a bit off guard today, since generally `process.env` values are strings that need to be coerced into other types. It would be nice to indicate in the docs that this is not the case for these values in the worker's env.

### DIFF
--- a/docs/worker-runtime-details.md
+++ b/docs/worker-runtime-details.md
@@ -4,8 +4,8 @@ In addition to any environment variables pre-configured for your worker via `wat
 
 Name | Description
 --- | ---
-Subject | the message's subject
-Message | the message's body
+Subject | the message's subject, parsed by `JSON.parse`
+Message | the message's body, parsed by `JSON.parse`
 MessageId | the message's ID defined by SQS
 SentTimestamp | the time the message was sent
 ApproximateFirstReceiveTimestamp | the time the message was first attempted


rclark:
This is surprising behavior to me, and I believe I'd consider it a bug, not a feature. Surprising because your worker process is a child process. The Watchbot application parses the subject and message in its process, then provides those values to the worker as environment variables on that child process. I don't understand how the child process could receive parsed objects/array.

jseppi:
I think I have something wrong here, but I'm not quite sure what yet...

jseppi:
Alright, I think what tripped me up is that when I write to the Watchbot SQS queue, I have to `JSON.stringify` the full SQS `MessageBody` and also `JSON.stringify` the inner `Message`:

```js
const params = {
  MessageBody: JSON.stringify({
    Subject: 'ddb-s3-record',
    Message: JSON.stringify(myJsonMessage),
  }),
  QueueUrl: QUEUE_URL,
};
await sqs.sendMessage(params).promise();
```

Then in my worker code, the `Subject` and `Message` have been set into `process.env` after being [`JSON.parse`'d](https://github.com/mapbox/ecs-watchbot/blob/4c51b69b95870ddfd48db42d008be5dd23a42568/lib/message.js#L42).

I do indeed need to do another `JSON.parse(process.env.Message)` to get back `myJsonMessage`.

Some example code or documentation would probably help since it took me a little while to get it figured out, but this PR doesn't address it properly.

PR #355, opened 2021-02-22
Node 12 support

hcourt:
Stacks created by ecs-watchbot are using Node 10, which is EOL April 30th.

This PR will upgrade ecs-watchbot Lambdas to use Node 12, and stop generation of Node 10 binaries in favor of Node 12.

cc/ @mapbox/data-platform for review, please


drboyer (approved):
LGTM, thanks for this change @hcourt!

PR #356, opened 2021-02-26
Use dockerhub credential for alpine binary builder

hcourt:
Make ecs-watchbot-generate-binaries use our dockerhub credential for the AlpineBundler CodeBuild project.  This PR doesn't also add the credential to the standard Bundler project because it uses a `codebuild` curated image, and AWS requires you use CodeBuild credentials for curated images.

Thanks to @rclark for pointing this out!

cc/ @mapbox/data-platform


drboyer (approved):
Thanks for this! I see, so Amazon will pull their own images from a non-Docker registry, so we don't hit rate limits with the standard bundler for that reason?

hcourt:
Yeah, that's my understanding.  Thanks!

PR #357, opened 2021-02-26
Use username/password formatted credentials for DockerHub

hcourt:
The `accesstoken` secret, which is used by both codebuild-helper and the generate-binaries pipeline, is not in the [expected format](https://docs.aws.amazon.com/codebuild/latest/userguide/sample-private-registry.html) used by the generate-binaries pipeline for the `RegistryCredential` attribute.  I created a similar secret that is in the expected format and stored it in `credentials`.

cc/ @mapbox/data-platform

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -6,7 +6,7 @@ phases:
       nodejs: 12
   pre_build:
     commands:
-      - aws secretsmanager get-secret-value --secret-id ${DOCKER_HUB_TOKEN_KEY} --region us-east-1 --query SecretString --output text | docker login --username mapboxmachinereadonly --password-stdin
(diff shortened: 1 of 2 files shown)


rclark:
How hard would it be to adjust the usage of the existing `/token` secret to use this `/credentials` one and parse the JSON structure? Having 2 secrets looks to me like its a rotation accident waiting to happen.

hcourt:
@rclark might be possible with the `:` syntax used to access a specific JSON key in the secret?  I'll try that over in the codebuild repo.

hcourt:
@rclark created a PR to do what you suggested with you as reviewer, will DM it to you.

hcourt:
Oh, nevermind you got to it, thanks!

rclark (approved):

PR #366, opened 2023-03-23
Update watchbox-progress

YannickMeeus:
**Why**

I have been pulling on CVE threads all week. Now it's ecs-watchbot's turn. I'm mostly interested in patching up dev-dependencies, but some runtime ones snuck in as well.

I've had to skip some tests, as they've been broken for well over a year now. I've pinged internally, and we'll see what happens with those. I'm not making the situation worse at least, which is something.


mapsam (approved):
LGTM

PR #378, opened 2023-09-07
Update to node 18

pratikyadav:
Update to node 18.

PR #380, opened 2023-09-21
Update to AWS SDK 3

mappingvermont:
As part of our [node 18 upgrade](https://github.com/mapbox/ecs-watchbot/pull/378) we bumped our lambdas to node 18.

The TotalMessagesLambda - [defined in lib/template.js](https://github.com/mapbox/ecs-watchbot/blob/944cc2c9242bcc52dc6f2e7c6f09fb08e055d7cc/lib/template.js#L782) - uses the aws-sdk-v2 package to poll SQS to understand how many watchbot messages are in the queue.

Previous node lambda runtimes had aws-sdk-v2 installed by default, but the node 18 now has aws-sdk-v3 installed, which has a different import. This PR updates all our code to use aws-sdk-v3 to ensure that that lambda works and also future proof our code as aws-sdk-v2 is likely entering maintenance mode soon.

While we have migrated to aws-sdk-v3, only the imports and a few function signatures have changed. This aws-sdk-v2 style code will still work with v3, but it isn't officially recommended. If another maintainer has time to fully migrate to the new 'command' style, that would be great.

In the meantime, this PR:
- Migrates from aws-sdk-v2 to v3
- Gets the TotalMessagesLambda working again
- Appropriately scales up and scales down watchbot jobs 
- Removes the aws-sdk-mock library used to mock aws-sdk v2 in favor of aws-sdk-client-mock to mock v3
- Updates all tests ✅ 

## Testing

To fully test, I first created a [tagged release](https://github.com/mapbox/ecs-watchbot/releases/tag/v9.0.1-dev1)  in the GH UI, then built them in docker locally:

```
docker build -t ecs-watchbot -f 'test/Dockerfile' .
mbx env -a default
docker run  $(env | grep 'AWS' | sed 's,^,-e ,') -it ecs-watchbot /bin/bash
AWS_REGION=us-east-1 CODEBUILD_RESOLVED_SOURCE_VERSION=<SHA of tagged version> node bin/watchbot-binary-generator.js
```

This build the binaries in my local docker, then uploaded them to:
```
s3://watchbot-binaries/linux/v9.0.1-dev1/watchbot
```

The proper way to do this would likely be to create a new codepipeline staging stack, but @oielbanna and I ran into lots of token / permissions / webhook issues with this. This approach gives us a pretty good approximation of the binaries. It's not perfect - the local image is based on `public.ecr.aws/docker/library/node` and the codepipeline image is `aws/codebuild/standard:7.0` - but it's good enough to test things. (We could build `aws/codebuild/standard:7.0` locally and then use that, but building that base image locally takes ~1 hour and at this point I was ready to test things 😬 ).

I tested the new binary by creating a mapbox-places watchbot stack with [this dockerfile](https://github.com/mapbox/mapbox-places/blob/4005e7637afe613a3922af7c50da2b7c66c19708/Dockerfile#L9-L11). Note the `ENV DUMMY` line to bust the cache and force re-downloading the binary after I pushed a new update to S3.

This binary worked as expected - building successfully _and_ shutting down the stack when no messages were present 🎉 

watchbot_scales_down

## Questions

I like the new mocking library, but there are a few things that seem suboptimal:

```js

 assert.equal(sqsMock.commandCalls(ChangeMessageVisibilityCommand, {
    QueueUrl: 'https://sqs.us-east-1.amazonaws.com/123456789012/fake',
    ReceiptHandle: 'a',
    VisibilityTimeout: 8
  }, true).length, 1, 'returns message to queue with backoff on visibility timeout');

```

^^ This seems a little heavy handed - I want to just check if SQS was called with those params. The mocking library returns an array from this check . . . and I then need to check its length. I'd love a way a shortcut to get it to return true / false. Not a big deal, but if anyone has experience with this library I'd take the help 🙏 

## Next steps

- [ ] PR review
- [ ] Merge
- [ ] New v9.0.1 release

cc/ @oielbanna @pratikyadav


oielbanna (approved):
Saving the day 🙏

pratikyadav (approved):
Thank you so much @mappingvermont 

Nitpick: can you add a changelog entry and update readme version reference from 9.0.0 -> 9.0.1?

Otherwise looks good. 👍

PR #384, opened 2023-11-15
Base Fargate construct

oielbanna:
## What changed?
This PR is part of a larger restructuring of this lib to use CDK, so if there are some code comments or missing things please disregard them as this isnt being merged to `main` just yet. This PR introduces `FargateWatchbot` construct which acts very similarly to the previous template provided. 
* Some props were renamed to fit the CDK name. For example `maxScalingCapacity` was previously `maxSize` I did this in order to make it easier for people to look at the CDK docs if needed. I also added in the code docs what the previous value was. This can also be added as extra migration docs if needed.

### Why did I 'borrow' `QueueProcessingFargateService` from CDK and make some changes to it?
Good question. Im writing this for our successors when they see this and get confused why did we just copy a whole bunch of code.
Some properties in CDK can only be provided when an object is getting created and can't be modified after. Namely `privileged`, `memoryReservationMiB`, & `readonlyRootFilesystem`. All 3 of these properties were being provided by the previous version of this library as params that can be changed and unfortunately they get locked down in CDK once you call `addContainer` to create a new `TaskDefinition` resource. 

## Whats next?
* Adding alarms
* Updating docs
* Creating Ec2Watchbot


lib/QueueProcessingFargateService.ts
```diff
@@ -0,0 +1,109 @@
...
+    public readonly service: FargateService;
+    /**
+     * The Fargate task definition in this construct.
+     */
+    public readonly taskDefinition: FargateTaskDefinition;
+
+    /**
+     * Constructs a new instance of the QueueProcessingFargateService class.
+     */
+    constructor(scope: Construct, id: string, props: MapboxQueueProcessingFargateServiceProps) {
```
jlouk, line 56: I remember seeing in `QueueProcessingFargateServiceProps` that the queue parameter was optional. Do we want to add a check to ensure that it is defined before spinning anything up? For our specific use case I'm thinking this needs to be set for it to be functional

lib/watchbot.ts
```diff
@@ -1,21 +1,375 @@
...
+  /**
+   * @default {prefix}-${stackName}
+   */
+  readonly containerName?: string;
+
+  /**
+   * The intervals for scaling based on the SQS queue's ApproximateNumberOfMessagesVisible metric.
+   * @see https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs_patterns.QueueProcessingFargateService.html#scalingsteps
+   */
+  readonly scalingSteps?: ScalingInterval[];
```
jlouk, line 31: Another one where I'm wondering if we want to make this mandatory? Or is there a default value for the scaling steps?

oielbanna: The default is defined in [the docs](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs_patterns.QueueProcessingFargateService.html#scalingsteps). I think its fine to keep it optional but we could change the default if you think its necessary. I know this is a bit different than the default value we currently have

jlouk: Sweet, the default looks good to me. If we want to change it to a custom default we can circle back later on and edit this

package.json
```diff
@@ -35,22 +35,22 @@
   "devDependencies": {
     "@mapbox/eslint-config-mapbox": "^2.0.1",
     "@mapbox/mock-aws-sdk-js": "^1.0.0",
+    "@types/jest": "^29.5.5",
+    "@types/node": "20.7.1",
+    "aws-cdk-lib": "2.102.0",
     "cli-spinner": "^0.2.10",
+    "constructs": "^10.0.0",
```
oielbanna, line 42: me thinks `constructs` and `aws-cdk-lib` should be part of `peerDependencies`, is that true?

lib/QueueProcessingFargateService.ts
YannickMeeus: Unless we need specific queue settings that are inherent to Mapbox, wouldn't it make sense for it to manage its own queue construct if not provided? Most L3 pattern constructs operate in this way, and it allows for quick prototyping.

package.json
YannickMeeus: I believe so, yes.

YannickMeeus: With a pretty wide version constraint, if possible, such that we allow consumers a degree of freedom instead of pinning them to a patch version.

lib/QueueProcessingFargateService.ts
oielbanna: The code you are commenting on is a construct will only be used internally by the lib and we are overriding the queue properties already but none of this is exposed to the user anyway. 

**The only queue properties that will be exposed to the user are `fifo`, `deadLetterThreshold` and `retentionPeriod`** as you can see in [WatchbotProps](https://github.com/mapbox/ecs-watchbot/blob/base-fargate-construct/lib/watchbot.ts#L21).  All other queue props get overridden depending on if the user specifies they want a fifo queue or not. Are you guys suggesting I change this behavior to either accept these 3 properties, or accept a queue? Here is the relevant code snippet https://github.com/mapbox/ecs-watchbot/blob/4b8773de0c82ee3501e4e239475d2ebbad062201/lib/watchbot.ts#L232-L249

package.json
oielbanna: Added https://github.com/mapbox/ecs-watchbot/blob/4241d33d50ce6f1307454c942db9805b30dd97ca/package.json#L69-L72

lib/QueueProcessingFargateService.ts
jlouk: >I can override it and make it a required prop if you prefer? Otherwise if we keep it optional, it would create a default queue if its not already provided. What would you rather we do?

I'm actually okay with keeping it optional if its creating a default queue. Might make for a more streamlined experience if the end user was looking to create a queue that also needed to be monitored for autoscaling etc.

lib/QueueProcessingFargateService.ts
```diff
@@ -0,0 +1,109 @@
...
+            memoryReservationMiB: props.memoryReservationMiB,
+            readonlyRootFilesystem: props.readonlyRootFilesystem,
+        });
+
+        // The desiredCount should be removed from the fargate service when the feature flag is removed.
+        const desiredCount = FeatureFlags.of(this).isEnabled(cx_api.ECS_REMOVE_DEFAULT_DESIRED_COUNT) ? undefined : this.minCapacity;
+
+        // Create a Fargate service with the previously defined Task Definition and configure
+        // autoscaling based on cpu utilization and number of messages visible in the SQS queue.
+        this.service = new FargateService(this, 'QueueProcessingFargateService', {
```
jlouk, line 87: Apologies if I'm not understanding this correctly - but don't we want to use the [QueueProcessingFargateService Construct](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs_patterns.QueueProcessingFargateService.html)

lib/watchbot.ts
```diff
@@ -1,21 +1,376 @@
...
+
+
+    this.deadLetterQueue = new Queue(this, 'DeadLetterQueue', {
+      fifo: this.props.fifo,
+      queueName: `${this.stack.stackName}-${this.prefixed('DeadLetterQueue')}${this.props.fifo ? '.fifo' : ''}`,
+      retentionPeriod: this.props.retentionPeriod || Duration.days(14),
+      contentBasedDeduplication: this.props.fifo,
+    });
+
+    this.queue = new Queue(this, 'Queue', {
```
jlouk, line 239: Would there ever be a use case where we the queue would be pre-existing and we want the watchbot to monitor it? I'm still wrapping my head around everything but _if_ that was the case, perhaps we could have an optional parameter for the queue? And if its not provided we create it like it is now?

lib/QueueProcessingFargateService.ts
oielbanna: No worries, it does look a bit odd! I explained my reasoning in the PR description but the TLDR is that the three properties privileged, memoryReservationMiB, & readonlyRootFilesystem cannot be mutated after creation in CDK. The `QueueProcessingFargateService` construct does not accept any of them as props so that would force us to use the default values provided by CDK. We cannot do that because it would break existing watchbot workflows that do pass those values in their params. I borrowed the code for this construct from aws-cdk-lib so it just extends the construct with extra functionality to pass on these 3 properties. Hopefully that makes sense 😄

lib/watchbot.ts
```diff
@@ -1,21 +1,376 @@
...
+      prefix,
+      containerName: `${prefix}-${this.stack.stackName}`,
+      structuredLogging: false,
+      readonlyRootFilesystem: true,
+      maxJobDuration: Duration.seconds(0),
+      family: props.serviceName,
+      cluster: Cluster.fromClusterAttributes(this, `${id}Cluster`, {
+        clusterName: `fargate-processing-${props.deploymentEnvironment}`,
+        vpc: Vpc.fromLookup(this, `${id}VPC`, {
+          vpcId: 'vpc-id'
```
oielbanna, line 349: @jlouk I just remembered I left this placeholder, what should this value be?

lib/QueueProcessingFargateService.ts
YannickMeeus: > three properties privileged, memoryReservationMiB, & readonlyRootFilesystem cannot be mutated after creation in CDK

That makes me wonder "why". Do we know whether the CDK construct authors are trying to bake in good behaviour and we're just taking a step back because that's how we've done it before?

lib/template.js
YannickMeeus, line 1: Will there be tests covering these additional resources?

YannickMeeus (approved):

PR #386, opened 2023-12-04
PLAT-709 Add Fargate Alarms

oielbanna:
## What changed?
* This PR uses cdk-monitoring-constructs to add the following alarms
  * WorkerErrors: measured errors in task execution based on log group metric filter 
  * CpiUtilization of the service
  * MemoryUtilization of the service
  * QueueSize if it gets too large
  * DeadLetterQueueSize if any messages exist
* The `monitoring` object is exposed as a public readonly class member in case users want to add additional alarms
* The properties object `alarms` forces users to have an alarm on this system while giving them more alarm configuration options.


package.json
```diff
@@ -55,6 +55,7 @@
     "typescript": "~5.2.2"
   },
   "dependencies": {
+    "@aws-cdk/aws-redshift-alpha": "^2.112.0-alpha.0",
```
oielbanna, line 58: This lib is a peer dependency for `cdk-monitoring-constructs`. Does that mean we also need to add it as a peer dependency here?

lib/watchbot.ts
```diff
@@ -204,6 +208,21 @@ export interface WatchbotProps {
...
 }
 
+export type WatchbotAlarms = {
+  action: ITopic;
+  memoryUtilization?: AlarmProps;
+  cpuUtilization?: AlarmProps;
+  queueSize?: AlarmProps;
+  dlqSize?: AlarmProps;
+  workersFailure?: AlarmProps;
+}
```
oielbanna, line 218: Users are required to pass an alarm `action` but all other properties are optional.

lib/watchbot.ts
```diff
@@ -346,7 +470,7 @@ export class FargateWatchbot extends Resource {
       cluster: Cluster.fromClusterAttributes(this, `${id}Cluster`, {
         clusterName: `fargate-processing-${props.deploymentEnvironment}`,
         vpc: Vpc.fromLookup(this, `${id}VPC`, {
-          vpcId: 'vpc-id'
+          vpcId: 'vpc-id' // TODO update
```
jlouk, line 473: Since we haven't solved the sharing data across accounts problem, I've typically created config files that would store the vpc id along with other necessary aws fun [here](https://github.com/mapbox/multi-account-ecs/blob/main/lib/config/aws.ts). We shouldn't have to change the vpc id's so this should be a low risk hard coding

package.json
jlouk: I'm no node wizard but wouldn't `cdk-monitoring-constructs` pull this package + all other necessary deps in when imported on line 63?

jlouk (approved):
LGTM. 2 quick comments - one about how to resolve for the vpc id - and another (more of a q than comment) about the peer dependencies. I don't think either are blockers, and the vpc id can be fixed in a future iteration

package.json
oielbanna: Not exactly, If you try to install `cdk-monitoring-constructs` without having the other packages, it will fail with a dependency error. Its added as a peer dependency to let you manage the version you need so ill think ill also do the same. I dont want `@aws-cdk/aws-redshift-alpha` to be bundled with my code

PR #388, opened 2023-12-06
PLAT-720 Progress table

oielbanna:
## What changed?

This PR adds the `reduce mode` feature. This dynamo table definition used to be created through `watchbot-progress` but I thought I would include it directly in this template instead of having to add another dependency.

❓  Open to feedback 
I added an extra property to the `WatchbotProps` interface but I'm not sure this is the best way to do this. I considered creating another construct that inherits `FargateWatchbot` or alternatively creating a helper function that users can invoke after instantiation. Open to other options here as well 

## Tests
Template tests are added for this resource


lib/watchbot.ts
```diff
@@ -490,6 +535,11 @@ export class FargateWatchbot extends Resource {
       fifo: false,
       deadLetterThreshold: 10,
       retentionPeriod: Duration.days(14),
+      reduceModeConfiguration: {
+        enabled: false,
```
jlouk, line 539: qq - does this code block do anything if `enabled` is set to false?

oielbanna: nope, it will be ignored

PR #392, opened 2024-01-16
Fix bug when using CDK to create FIFO queues

mappingvermont:
Apparently there's a known issue with creating FIFO queues using CDK: https://github.com/aws/aws-cdk/issues/8550

This workaround worked for me - not sure if it's best for all but ticketing out here 🙏


oielbanna (approved):
Thank you for adding this!!

